PR #59, opened 2015-04-07
Add configuration option for custom intercept logic.

kkaehler:
I really want to integrate silk into our dev/staging environments but we need more fine-grained controls over when it is enabled. Let me know if you've been thinking of a better way to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,9 +353,26 @@ Note that in the above screenshot, this means that the request took 29ms (22ms f
 
 On high-load sites it may be helpful to only record a fraction of the requests that are made.To do this add the following to your `settings.py`:
 
+Note: This setting is mutually exclusive with SILKY_INTERCEPT_FUNC.
+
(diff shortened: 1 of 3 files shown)


mtford90:
I think this is a great way of doing that, slots in nicely with everything else. Cheers!

kkaehler:
Unfortunately I screwed up the implementation of this.

Since I did this by passing a reference to the function this gets set when settings are imported and before  django has added the application name to the path. What this means is that you get import errors if you want to import any application specific code for this method.

I have a fix in which you pass a string referencing the file instead of the function itself (this is what the debug-toolbar does). Would you prefer for me to remove the pass by reference option or leave both?

mtford90:
@kkaehler woops :).

Yeah prob best to remove the pass by ref and make sure this is all documented rather than leave something in that could leave people confused ;)

PR #62, opened 2015-04-13
Update model_factory.py

karabijavad:
just in case the json body is bad json, this prevents the silk middleware from blowing up when the application code may want to get see the response


karabijavad:
for what its worth, i have a unit test which was specifically providing bad json in the request body, but my application code was not reached because silk was blowing up at this point

mtford90:
Awesome, good catch, cheers

PR #75, opened 2015-08-27
Support UUID in request headers

rodcloutier:
Django support inserting headers fields of type UUID. 

This pull request enabled the serialization of the UUID type when encoding the headers.


mtford90:
Cheers!

PR #76, opened 2015-09-01
Fixed report handling timing not included in meta-timing

rodcloutier:
This pull request aim to fix a problem where the timing for processing the request information was not taken in account in the meta-timing. 
The save was done in a meta-timing context.


mtford90:
Good catch, cheers - sorry for the delay on getting to this.

PR #77, opened 2015-09-15
Added the use of Lambdas in settings.py to the README.

bryson:
Also included a bit of whitespace cleanup that happened automatically in my editor: removed trailing spaces; converted a couple places with mixed space/tab indentation to only 4-space tabs.


mtford90:
Cheers!

PR #81, opened 2015-10-14
Namespace templatetags so they don't clash with existing application templatetags

lastkarrde:
Prefix template tags with silk_ to prevent them clashing with application template tags.


mtford90:
awesome, cheers

Issue #85, opened 2015-10-30
UnicodeDecodeError in sql.py: leads to 500 internal error


ChristopherRabotin:
It seems that silk does not support UTF-8 characters in SQL queries.

```
Internal Server Error: /
Traceback (most recent call last):
  File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py", line 223, in get_response
    response = middleware_method(request, response)
  File "/app/.heroku/python/lib/python2.7/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py", line 331, in wrapper
    return wrapped(*args, **kwargs)
  File "/app/.heroku/python/lib/python2.7/site-packages/silk/middleware.py", line 123, in process_response
    self._process_response(response)
  File "/app/.heroku/python/lib/python2.7/site-packages/silk/middleware.py", line 107, in _process_response
    silk_response.save()
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/base.py", line 827, in _save_table
    forced_update)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/base.py", line 877, in _do_update
    return filtered._update(values) > 0
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/query.py", line 580, in _update
    return query.get_compiler(self.db).execute_sql(CURSOR)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 1062, in execute_sql
    cursor = super(SQLUpdateCompiler, self).execute_sql(result_type)
  File "/app/.heroku/python/lib/python2.7/site-packages/silk/sql.py", line 38, in execute_sql
    sql_query = q % params
UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 1: ordinal not in range(128)
```

trik:
as a temporary fix you can change model_factory.py:261 from
`silky_response.raw_body = content`
to
`silky_response.raw_body = content.encode('UTF-8', 'replace')`

avelis:
[...]

PR #94, opened 2016-01-19
Response bodies are now stored b64 encoded (support for binary responses).

trik:
removed some useless encoding/decoding of response body (why it was decoded from original charset before json dump and load??)

### DIFF
--- a/silk/model_factory.py
+++ b/silk/model_factory.py
@@ -2,6 +2,7 @@ import json
 import logging
 import sys
 import traceback
+import base64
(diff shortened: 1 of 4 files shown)


avelis:
@trik Thanks for making this effort and resolving #1 & #85 I know it was a long standing pain point for awhile.

trik:
@avelis don't mind! i like this project (and i'm using it for work) so i think i'll make some more bug fixing / enhancing

PR #99, opened 2016-01-27
URL patterns are just Python lists for Django 1.9+

blag:
See the deprecation notes here: https://docs.djangoproject.com/en/1.9/ref/urls/#patterns

### DIFF
--- a/silk/urls.py
+++ b/silk/urls.py
@@ -1,4 +1,4 @@
-from django.conf.urls import patterns, url
+from django.conf.urls import url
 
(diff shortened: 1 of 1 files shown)


blag:
I should note that this is not compatible with Django < 1.8, but it should work with Django 1.8 just fine.

avelis:
Sorry for the delay @blag I was on vacation.

Thank you for contributing this! :tada:

blag:
No worries, thanks for making Silk! I've already used it to diagnose a problem I was having (and this PR silences one of the many warnings I was getting).

PR #100, opened 2016-02-03
Django 1.9 compatibility for the csrf context processor

blag:


avelis:
@blag Much appreciation again for adding more Django 1.9 compatibility.

PR #104, opened 2016-03-12
Fixing Depreciation, Saving and Performance Tweaks

Wrhector:
I can break this up, but simply this fixes:

- Django deprecation warnings.
- A previous commit broke sql and profile saving entirely.
- Stop attaching the profiler every run if not used.
- Better update of total queries (Though this really could be refactored into one update)

### DIFF
--- a/project/tests/test_config_meta.py
+++ b/project/tests/test_config_meta.py
@@ -23,7 +23,7 @@ class TestConfigMeta(TestCase):
         delete_all_models(Request)
         DataCollector().configure(Request.objects.create())
         response = self._mock_response()
-        SilkyMiddleware()._process_response(response)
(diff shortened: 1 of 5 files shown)


avelis:
@Wrhector Thanks for fixing the broken collection. Truly was a mistake on my part. Also, I appreciate the effort to the project.

In the interest of progress I have decided to to merge it in!

wm3ndez:
@Wrhector 958d843 breaks django=1.7.x, is this version of django no longer supported?

Wrhector:
Django 1.7 has been end of life for 6 months now, though oddly travis passed checks with Django 1.7, I'll write a patch to support both.

wm3ndez:
@Wrhector I already sent a patch, thanks

PR #107, opened 2016-03-30
Supporting django<1.8

wm3ndez:


avelis:
@wm3ndez Thanks for the contribution.

PR #110, opened 2016-05-25
remove south migrations as not needed in less then 1.7

auvipy:


avelis:
@auvipy Before I consider merging this in would this change drop Django 1.6 support? I ask that because if it indeed does then the merge request needs to reflect that expectation.

auvipy:
mmm this is one of many 1.6 feature drop :) I believe sending small changes at a time but if you have other suggestions I would love to embrace that

avelis:
@auvipy Small changes at a time works for me.

PR #111, opened 2016-05-29
Render function instead of render_to_response

auvipy:
this is the preferred one

### DIFF
--- a/project/example_app/views.py
+++ b/project/example_app/views.py
@@ -1,7 +1,7 @@
 from time import sleep
 
 # Create your views here.
-from django.shortcuts import render_to_response
+from django.shortcuts import render
 from example_app import models
 from silk.profiling.profiler import silk_profile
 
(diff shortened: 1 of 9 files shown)


avelis:
@auvipy After [reading the following stackoverflow question](http://stackoverflow.com/questions/5154358/django-what-is-the-difference-between-render-render-to-response-and-direc) I am ok with merging this in. If you could resolve the failing unit tests In this PR that would be greatly appreciated.

auvipy:
I will try to fix the tests failing first then.

avelis:
@auvipy I appreciate the effort. I also understand that there might be a failing test not associated to this change. I can be reasonable and expect that test to not get addressed in this PR.

auvipy:
@avelis failed tests are a lot which take time to fix. so i would suggest to merge for now as is. i'm promising to comeback soon with more patches.

avelis:
@auvipy I have merged in this commit.

PR #118, opened 2016-07-04
remove claim to support django 1.6

chrono:
it's confusing to look at master's readme claiming support and then noticing
it's not supported.


avelis:
@chrono Thank you for your contribution.

PR #123, opened 2016-07-08
Give users the ability to export .prof binary files for every request

hanleyhansen:
## What
Give the user the option the generate the binary `.prof` data and expose it via a download button in the profiling page. This functionality is driven by a new setting called `SILKY_PYTHON_PROFILER_BINARY`. Docs have been updated to reflect it's usage.

The download button will appear on the right only if there is an associated `.prof` file with the request

screen shot 2016-07-08 at 11 54 01 am

## Why
`silk` produces beautiful human readable output on it's profiling page via `pstats` but there was no way to extract this data to use in other programs like `snakeviz`. This PR enables the user to extract said binary data.


avelis:
@hanleyhansen Since you added a field to the Request model. Is it possible for you to add the migration file for that model change.

avelis:
@hanleyhansen Thank you for this feature and general contribution to this repo.

PR #128, opened 2016-07-13
use any readme

SzySteve:
#### What:
* Include any readme file

#### Why:
* See issue: https://github.com/django-silk/silk/issues/127
* The markdown README file is not included in the wheel, so systems without pyandoc cannot `pip install django-silk`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 include LICENSE
-include README.rst
+include README*
 recursive-include silk/templates *
 recursive-include silk/static *
 recursive-include silk/code_generation/ *.py


avelis:
@hanleyhansen @SzySteve That was exactly what I was going to do. 👍

hanleyhansen:
@avelis do you mind if we simplify and just keep the `md`? then we wouldn't need `pypandoc`

avelis:
@hanleyhansen Correct me if I am wrong but Pypi only accepts rst based descriptions when publishing releases. This is the main reason I have the conversion logic in there.

hanleyhansen:
@avelis you are correct. makes sense now. so only the machine uploading to pypi needs the pypandoc dependency. so this should work.

avelis:
@hanleyhansen I will work on a maintenance release with the manifest change shortly.

I appreciate the PR to fix this install issue.

hanleyhansen:
@avelis awesome thanks!

Issue #26, opened 2014-07-17
IntegrityError: duplicate key value violates unique constraint "silk_response_request_id_key"


synotna:
After activating silk certain urls began erring with:

IntegrityError: duplicate key value violates unique constraint "silk_response_request_id_key"
DETAIL:  Key (request_id)=(1166) already exists.

Sentry stack trace: http://toolbox1.tedc.de:9000/bidev/esldj/group/131/

miguelfg:
Hi,

Same problem here :-( 
Here the trace in case it helps:

Traceback (most recent call last):

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 199, in get_response
    response = middleware_method(request, response)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/silk/middleware.py", line 105, in process_response
    self._process_response(response)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/silk/middleware.py", line 94, in _process_response
    silk_response = ResponseModelFactory(response).construct_response_model()

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/silk/model_factory.py", line 246, in construct_response_model
    body=b)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/manager.py", line 157, in create
    return self.get_queryset().create(**kwargs)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/query.py", line 322, in create
    obj.save(force_insert=True, using=self.db)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/base.py", line 545, in save
    force_update=force_update, update_fields=update_fields)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/base.py", line 573, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/base.py", line 654, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)

  File "/home/project_name/.virtualenvs/project_name/local/lib/python2.7/site-packages/django/db/models/base.py", line 687, in _do_insert
    using=using, raw=raw)

[...]

PR #129, opened 2016-07-13
don't crash when a route is 404

chrono:
multiple issues:
 * resolve() might throw, handle that case
 * unhandled exception in `process_request` leads to last successful
   `request_model` remain in `DataCollector` and being used to create
   a response model for the current failed request, leading to
   a constraint violation thrown by the db (creating another response
   for the previously succesful request)

resolution
 * handle 404 by saving NULL/None as view name
 * clean out `DataCollector` state when starting a request, before doing
   anything.

### DIFF
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -85,18 +85,19 @@ class SilkyMiddleware(object):
 
     @silk_meta_profiler()
     def process_request(self, request):
-        request_model = None
-        if _should_intercept(request):
-            Logger.debug('process_request')
-            request.silk_is_intercepted = True
-            self._apply_dynamic_mappings()
-            if not hasattr(SQLCompiler, '_execute_sql'):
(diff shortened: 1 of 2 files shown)


avelis:
@chrono This might indirectly help #26

avelis:
@chrono Thanks for the contribution and for addressing this.

toSterr:
Hi, when is it planned to deploy this fix as a new version of silk?

avelis:
@toSterr I will try to get to by EOD today.

toSterr:
@avelis thanks for the response, since there is no new version I would like to ask if it is possible to do it before Friday EOB?

avelis:
@toSterr I think I can do that. Sorry for the delays on release.

avelis:
@toSterr I published 0.6.2 to PyPi. I also made a release for it on Github for reference.

PR #133, opened 2016-08-03
Namespaced loggers

aehlke:
Enables finer-grained logging configuration. If I want to suppress e.g. https://github.com/django-silk/silk/blob/6f8a5741115640dc752ef95951058b420d55dbe8/silk/profiling/profiler.py#L106 but not other warnings, with this patch I can change the logging level for `"silk.profiling.profiler" independently.

### DIFF
--- a/silk/collector.py
+++ b/silk/collector.py
@@ -18,7 +18,7 @@ TYP_SILK_QUERIES = 'silk_queries'
 TYP_PROFILES = 'profiles'
 TYP_QUERIES = 'queries'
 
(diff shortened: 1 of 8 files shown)


avelis:
@aehlke Thanks for the contribution. Could you add some documentation on the different types of logger namespaces on the README?

I know the build is failing but it is for unrelated reasons. Given that knowledge I have merged this in.

aehlke:
@avelis I have a couple other PRs I need to make first but I'll look into README!

PR #138, opened 2016-09-21
Django 1.10 compatibility

shanx:
Heya!

I made the SilkyMiddleware compatible with the new Django 1.10 middleware and also fixed the broken tests (typo with content_type not being a field on Response).

Thanks!
Remco


shanx:
Don't know why that specific test is failing in that specific deploy :( any ideas?

avelis:
@shanx That unit test might be inconsistent with when it breaks. It is something I had on the back of my mind to address. Either way I have merged your pull request.

Thanks again for the contribution!

PR #139, opened 2016-09-21
Remove trailing slashes in MANIFEST.in

leifdenby:
Trailing slashes breaks compilation on Windows. This commit removes these trailing slashes.


avelis:
@leifdenby Thanks for the contribution

PR #140, opened 2016-09-27
Operational Error When Silk Is Used On Big SQL Queries

hanleyhansen:
## What
Keep trying the `_process_response` until it works. It tends to fail with big MySQL queries.

## Why
MySQLdb specifically does not manage the connections and will error if you try to reuse a stale connection. One that has closed or expired. You need to explicitly check for this error case and handle it. This PR excepts the `OperationalError` whenever the `silk` middleware has to processs a relatively large query in it's `_process_response`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ Pillow>=3.2
 django>=1.8
 freezegun>=0.3,<0.4
 factory-boy==2.6.0
+mysqlclient==1.3.7
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ setup(
         'sqlparse',
         'Jinja2',
         'autopep8',
-        'pytz'
+        'pytz',
+        'mysqlclient'
     ]
 )
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -1,6 +1,8 @@
 import logging
 import random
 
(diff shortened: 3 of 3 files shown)


hanleyhansen:
@avelis we can discuss on where I put the `try` `catch` and maybe think about moving it somewhere closer to the actual cause so we don't mask any other errors. Let me know your thoughts.

avelis:
@hanleyhansen Could there ever be a scenario where you could get forever stuck in the while loop? I will likely merge but wanted to poke any holes at it before I did.

I terms of placement. I think it's fine.

hanleyhansen:
@avelis straight from the MySQL docs

> This exception is raised for errors which are related to MySQL's operations. For example: too many connections; a host name could not be resolved; bad handshake; server is shutting down, communication errors.

So the only way we can get stuck in a loop forever is if the DB server goes down. At the point the person has a different problem though. Too many connections will keep trying and tell there are some available connections. Communication errors, same thing. Other stuff will break before `silk` gets stucked in a loop forever. Otherwise just catching `OperationalError` and `AttributeError` should be safe within the `while True:` loop.

avelis:
@hanleyhansen Thanks for the contribution and providing a solution for this scenario!

Issue #142, opened 2016-10-02
Bad requirements for postgres based installations


hassonofer:
After merging #140, `pip install django-silk` fails for postgres based installations.

mysqlclient was added as a requirement, which produces the following error (if no mysql installation exist).

``` shell
Collecting mysqlclient (from django-silk)
  Using cached mysqlclient-1.3.9.tar.gz
    Complete output from command python setup.py egg_info:
    sh: mysql_config: command not found
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-KYR3WJ/mysqlclient/setup.py", line 17, in <module>
        metadata, options = get_config()
      File "setup_posix.py", line 44, in get_config
        libs = mysql_config("libs_r")
      File "setup_posix.py", line 26, in mysql_config
        raise EnvironmentError("%s not found" % (mysql_config.path,))
    EnvironmentError: mysql_config not found
```

avelis:
@hassonofer I had a feeling this was going to happen. For now install version `1.7.0`. I will have to figure out an SQL agnostic solution for what @hanleyhanson was looking for in the last release.

hassonofer:
First of all, thank you very much for the prompt response.

I did not test it myself, but as far as I know ```django.db.DatabaseError``` or the more general ```django.db.Error``` should catch this error.
As documented here: https://github.com/django/django/blob/149ace94dfc10504a0e69462c7737f5ce05340a4/docs/releases/1.4.txt#L880

hanleyhansen:
Hi @hassonofer,

I'll be home on 30 mins and will try that. It if works I'll put another PR up. I'm sore @avelis will merge it right away so we don't hold you up. 

Sorry for the inconvenience.

hassonofer:
There is not rush at all, I updated my puppet manifests today to stay on 0.7.0 :)

avelis:
@hassonofer Thanks for suggesting a `django.db.DatabaseError` error catch instead. Sounds like the right type of error to catch. @hanleyhansen If you get a PR working I will promptly merge as well.

I will also into taking down 1.7.1 from PyPi to avoid others from getting caught into this issue.

avelis:
@hanleyhansen I made a [branch](https://github.com/django-silk/silk/tree/database-error-try-catch) to try out a different class in the exception catching. Can you confirm/deny if that works for you?

avelis:
[...]

PR #144, opened 2016-10-05
Updates exception handling to use Django DatabaseError class

hanleyhansen:
Addresses #142

Using Django’s DatabaseError class as the exception to catch keeps the
library and this try catch database platform agnostic. That way
developers who have a Postgres or MySQL backend should mutually benefit
from this retry.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,3 @@ Pillow>=3.2
 django>=1.8
 freezegun>=0.3,<0.4
 factory-boy==2.6.0
-mysqlclient==1.3.7
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,5 @@ setup(
         'Jinja2',
         'autopep8',
         'pytz',
-        'mysqlclient'
     ]
 )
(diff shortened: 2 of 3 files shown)


hanleyhansen:
@avelis code changes look perfect. Exactly what I was going to do. Travis-CI is complaining about that one test for the 3.4 environment. Same intermittent test that has failed for us before:

```bash
FAIL: test_time_spent_db_with_path (tests.test_view_requests.TestGetObjects)
----------------------------------------------------------------------
  File "/home/travis/build/django-silk/silk/project/tests/test_view_requests.py", line 98, in test_time_spent_db_with_path
    self.assertTrue(num_results)
AssertionError: 0 is not true
```

So I think we can ignore that. Unfortunately I'm at the car dealership right now so I only have my work laptop on me but when I get home I can test this branch against the code that was causing the error and report back.

Up to you if you want to wait for that or if you wanna just go ahead and merge. Everything looks good to me.

avelis:
@hanleyhansen I have gone ahead & merged. Thanks for checking on the pull request for me.

PR #145, opened 2016-10-06
Travis config to test on the different django database backends.

mattjegan:
Hi all, I have added some config to `.travis.yml` to allow for testing on "most" of the Django database backends. I have included PostgreSQL, MySQL and SQLite3. The only default Django backend that is missing is Oracle. This can be included however it would mean pulling in [this project](https://github.com/cbandy/travis-oracle) and also a maintainer setting up an oracle account.

The motivation for this change is issue #142 and to allow for earlier detection so that a broken release doesn't make it to PyPi.

As you can [see on Travis](https://travis-ci.org/mattjegan/silk/builds/165481576) it works well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,23 @@ python:
     - "3.4"
     - "3.5"
 env:
-    - DJANGO="Django>=1.7.0,<1.8.0"
-    - DJANGO="Django>=1.8.0,<1.9.0"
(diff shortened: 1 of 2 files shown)


avelis:
@mattjegan Thank you for the contribution! Definitely a testing proactive approach I can get behind.

PR #148, opened 2016-11-16
Allow using Django 1.10 MIDDLEWARE setting instead of MIDDLEWARE_CLASSES

lockie:
Hello!
When using MIDDLEWARE setting from Django 1.10 (instead of MIDDLEWARE_CLASSES, like [README](https://github.com/django-silk/silk/blob/85af6e2d25cf1e1be44d5de142fdfa8ec11754e5/README.md) suggests), silk keeps complaining that
```
Cannot execute silk_profile as silk is not installed correctly.
```
This PR fixes that.

### DIFF
--- a/silk/profiling/profiler.py
+++ b/silk/profiling/profiler.py
@@ -3,6 +3,7 @@ import logging
 import time
 import traceback
 
+import django
 from django.conf import settings
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


avelis:
@lockie Thanks for the contribution!

lockie:
@avelis You're very welcome :)

PR #164, opened 2017-03-09
Fixes #161

### LINKED ISSUE #161: Silk shows 0 time for all requests?
I'm just trying silk for the first time and it looks awesome!  After adding the app and the middleware, and running the migrations, I'm seeing requests being logged, but I'm not seeing any time recorded.

2017-02-18 09_19_39-silky

I must be missing something.  Any idea?

PeRDy:
The middleware wasn't storing anything because when tried to disable python profiler and it doesn't exists (initialized with None) raise a ValueError, so that the middleware stops at this point and don't register anything.

### DIFF
--- a/silk/collector.py
+++ b/silk/collector.py
@@ -133,11 +133,11 @@ class DataCollector(with_metaclass(Singleton, object)):
             self.request.save()
 
     def stop_python_profiler(self):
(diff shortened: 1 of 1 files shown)


avelis:
Thanks @PeRDy for the contribution and addressing #161

PR #167, opened 2017-03-13
Improved MySQL support

### LINKED ISSUE #38: Failed to install index for silk.Request model: (1071, 'Specified key was too long; max key length is 767 bytes')
I get this warning when run    "python manage.py syncdb".

smaccona:
Adds support for `utf8` and `utf8mb4` character sets in MySQL's default
configuration.

Closes #38.


smaccona:
@avelis if you can review when you get a chance. Here are the 3 scenarios I tested for locally:

- Fresh install: works
- Upgrade over existing install on proper MySQL configuration with long index support: works, does not change underlying fields (they stay at width 300)
- Upgrade over modified install on MySQL with no long index support (user manually changed the field width or created a truncated index and faked the initial migration): works, does not change underlying field (they stay at whatever width the user had them)

avelis:
@smaccona Thanks again for working on this PR and long standing issue. Much appreciation from me.

I am planning a release with a major version bump. I will make sure to set the expectation this will break current installations. However, the upgrade scenario's you noted might work for others and I will highlight them.

PR #171, opened 2017-03-15
Fixes #170

### LINKED ISSUE #170: There is no reference to download a profile
README mentioned that exists the possibility of download a profile directly from the interface, but at the moment the button to download it as well as the profiling file content itself doesn't appear in profile pages.

PeRDy:
With these changes, the profile content will appear in profile detail view, as well as a link to download the profile file. I tried to reuse CSS classes, but the aspect can be improved.

### DIFF
--- a/silk/templates/silk/profile_detail.html
+++ b/silk/templates/silk/profile_detail.html
@@ -20,6 +20,22 @@
             font-size: 13px;
 
         }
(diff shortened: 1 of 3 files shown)


avelis:
@PeRDy Thanks for this contribution!

PR #176, opened 2017-05-09
Add text-align property to pyprofile class for readability

jeffreyckchau:
Hi, 

This PR adds the `text-align: left` to the pyprofile class in the profiling view. 

image

As shown in the above screenshot, without a text align the output is difficult to read with longer function names. 

image

With the `text-align` added, it is much more readable. 

Thank you for the consideration! This is an incredibly useful package for us.

### DIFF
--- a/silk/templates/silk/profile_detail.html
+++ b/silk/templates/silk/profile_detail.html
@@ -21,6 +21,10 @@
 
         }
 
+        .pyprofile {
+            text-align: left;
+        }
+
         a {
             color: #45ADA8;
         }


jeffreyckchau:
I'm having trouble interpreting the errors with the build here... 

For django==1.9.0 && python==3.5, silk cannot be installed correctly? How does that result from a static file change?

avelis:
@jeffreyckchau Thanks for the contribution. That will greatly increase the readability of that view. Much appreciated.

As for the failing build. There is a one test on one environment configuration that is failing. I have not gotten around to fixing it. But I don't impede contributions for the time being.

jeffreyckchau:
@avelis Ah, got it. Glad to help!

PR #177, opened 2017-05-26
Deprecation: update to warning

lammertw:
The 'warn' method is deprecated, use 'warning' instead


avelis:
@lammertw Thanks for addressing the deprecation warning!

PR #186, opened 2017-07-04
Replace django-silk organization with jazzband

albertyw:
This replaces a few urls which are incorrect now that the repo has been moved.  It also removes the outdated `README.rst` since `setup.py` is currently reading `README.md` and converting it to rst for pypi.

This may need some help in moving travis and gitter projects to point to the new repository.


avelis:
@albertyw Thanks for the contribution!

PR #188, opened 2017-07-04
Add explicit on_delete to foreign key and one to one relationships

### LINKED ISSUE #183: RemovedInDjango20Warning: on_delete will be a required arg for OneToOneField in Django 2.0.
Getting this warning with Django 1.11.1
I'm guessing if nothing else we should set on_delete to CASCADE as they suggest

/site-packages/silk/migrations/0001_initial.py:63: 
RemovedInDjango20Warning: 
on_delete will be a required arg for OneToOneField in Django 2.0.
Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior.
See [https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete)
('request', models.OneToOneField(to='silk.Request', related_name='response')),

/site-packages/silk/migrations/0001_initial.py:75:
RemovedInDjango20Warning: 
on_delete will be a required arg for ForeignKey in Django 2.0.
Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior.
See [https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete)
('request', models.ForeignKey(to='silk.Request', blank=True, null=True, related_name='queries')),

albertyw:
Fixes #183

### DIFF
--- a/silk/models.py
+++ b/silk/models.py
@@ -132,7 +132,10 @@ class Request(models.Model):
 
 class Response(models.Model):
     id = CharField(max_length=36, default=uuid4, primary_key=True)
-    request = OneToOneField(Request, related_name='response', db_index=True)
+    request = OneToOneField(
+        Request, related_name='response', db_index=True,
+        on_delete=models.CASCADE,
+    )
     status_code = IntegerField()
     raw_body = TextField(blank=True, default='')
     body = TextField(blank=True, default='')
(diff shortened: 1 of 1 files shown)


avelis:
@albertyw thanks again for all the PRs they really help this library.

PR #191, opened 2017-07-10
Added status code filter

danielbradburn:
I added a new user interface element to filter the requests based on http status code.


avelis:
@danielbradburn Thanks for the contribution of a status code filtering. I am positive it will increase the utility of the UI. 👍

PR #192, opened 2017-07-15
Visualize profile result

danielbradburn:
I added a visualization of the cprofile result as a graph in the profile detail page. This allows you to very quickly see the paths which are causing bottlenecks, for example:

image


avelis:
@danielbradburn Awesome contribution! When possible, could you add image you attached in the PR to the README.md file?

danielbradburn:
@avelis Ah, yes sorry I forgot about that. It's probably worth adding a note about the filtering and sorting changes I made to. Will make a separate PR with an updated README.

PR #193, opened 2017-07-15
Sortable sql table

danielbradburn:
Use sortable.js to make the sql queries table sortable on the queries page.


avelis:
@danielbradburn Thanks for the contribution!

PR #194, opened 2017-07-16
Added UI element to filter requests by http verb

danielbradburn:


avelis:
@danielbradburn Thanks again for the contribution! A UI method filter is a great addition here.

PR #197, opened 2017-07-17
Add missing gprof2dot to setup.py

danielbradburn:
This is required for graph visualization


avelis:
@danielbradburn Thanks for adding the library dependency. I am positive it would help on the next release

PR #199, opened 2017-07-18
left align pre tag text

smcoll:
Profller output, in a `pre` tag, is centered, due to an inherited rule for `#query-div` in `summary.css`:

```
#query-div {
    margin: auto;
    width: 960px;
    text-align: center;
}
```


avelis:
@smcoll I am rerunning the CI build before merging. I believe TravisCI is in a transition to Ubuntu Trusty and some builds failed because of it.

avelis:
@smcoll Thanks for the CSS enhancement!

Issue #33, opened 2014-08-17
Organise cProfile output as a sortable, more organised table.


mtford90:
It's currently just text dumped into a code block. Would be nice if this could be organised as a HTML table and possibly change the sorting etc (i.e. only sorted by cumulative atm)

danielbradburn:
I think the changes in https://github.com/jazzband/silk/pull/200 should fix this, can it be closed now?

avelis:
@danielbradburn Yes It can.

PR #200, opened 2017-07-23
#33 organise cprofile output as a sortable table

danielbradburn:
This fixes issues https://github.com/jazzband/silk/issues/33


avelis:
@danielbradburn Thanks for adding this feature!

PR #203, opened 2017-07-31
increase Request.prof_file max_length to 300

### LINKED ISSUE #201: Saving to Request.prof_file raises DataError for paths longer than 59 characters
The decorator is raising a DataError when it tires to save a `.prof` for any path greater than 59 characters, since the `Request.prof_file` is 100 characters by default, and the filename itself is 41 characters.

```
django.db.utils.DataError: value too long for type character varying(100)
```

In my case, the project path for local development is more than 60 characters, which means the file is saving to disk, but no profile data is being saved to the database.

smcoll:
references #201


smcoll:
i guess that technically, this doesn't fix the underlying issue of #201, but rather kicks the can down the road.   It's still possible that a DataError would be raised, and the profiler results not saved to the db, if the path is too long.  And it's not immediately apparent from the traceback why that is.

avelis:
@smcoll Could you add a small test for this change when you have a moment? I saw the failure and it's fine. It's just a precision error.

smcoll:
@avelis a test which asserts that the path can be up to a certain length?

avelis:
@smcoll A test can assert the length or the failure if going past 300. Nothing to fancy

smcoll:
@avelis what is the ideal behavior if the path is too long, or if the file can't be saved?

avelis:
@smcoll That raises a good question.
A couple options:

1. Graceful handling when a DatabaseError does occur.
2. If not able to gracefully handle at least an informative traceback to help users know why it failed.

However, I feel that introduces new/different scope into this PR. If you prefer, I can merge it in as is without issue.

Thoughts?

smcoll:
@avelis that seems reasonable to me.  The underlying issue about how to handle a file storage failure is not addressed in this PR, so i changed the description to relate to but not fix the issue.  A PR which actually changes the behavior of the app to handle storage failures seems like a better candidate for new tests.

avelis:
@smcoll Thanks for the contribution! I appreciate it.

PR #204, opened 2017-08-01
configurable storage class

### LINKED ISSUE #202: Storage class as a setting
It would make sense to me that the file storage class could be selected as a configuration, especially since many deployments these day are on ephemeral file systems.

We could set `storage`, something like this (untested):

```
from django.core.files.storage import get_storage_class
from django.conf import settings

storage_class = getattr(settings, 'SILKY_STORAGE_CLASS', 'silk.models.ProfilerResultStorage')
silky_storage = get_storage_class(storage_class)()

class Request(models.Model):
    # [...]
    prof_file = FileField(null=True, storage=silky_storage)
```

smcoll:
fixes #202

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,10 +169,10 @@ When enabled a graph visualisation generated using [gprof2dot](https://github.co
 A custom storage class can be used for the saved the generated binary `.prof` files:
  
 ```python
-SILKY_STORAGE_CLASS = 'path.to.StorageClass`
+SILKY_STORAGE_CLASS = 'path.to.StorageClass'
 ```
 
-The default storage class is `silk.models.ProfilerResultStorage`, and when using that you can specify a path of your choosing. You must ensure the specified directory exists.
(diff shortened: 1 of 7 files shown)


smcoll:
Rebased.  Note that the docs are updated and i've moved the default storage class to a new location.

avelis:
@smcoll Since this is introducing the concept of a customize-able storage. I am wondering if makes sense to add some config level tests asserting default behavior.

Thoughts?

smcoll:
Yes, i agree.  i'll add a test asserting the default storage class is used when the setting is not specified.  Would that be sufficient?

avelis:
@smcoll Yes that would be sufficient!

smcoll:
@avelis added- does that work?

avelis:
@smcoll Yes. Works for me. Thanks again for adding the test.

PR #209, opened 2017-08-08
support storage backends without path access

### LINKED ISSUE #208: storage backends without path access not supported
If an S3 storage backend, for example, is used as `SILKY_STORAGE_CLASS`, a `NotImplementedError` is raised when trying to save the file.

```
silk.profiling.profiler [ERROR] [profiler:__exit__] Exception when performing meta profiling, dumping trace below
  File "/home/myenv/lib/python3.5/site-packages/silk/middleware.py", line 117, in _process_response
    collector.finalise()
  File "/home/myenv/lib.python3.5/site-packages/silk/collector.py", line 156, in finalise
    with open(self.request.prof_file.storage.path(file_name), 'w+b') as f:
  File "/home/me/.pyenv/versions/evidencecare3/lib/python3.5/site-packages/django/core/files/storage.py", line 111, in path
    raise NotImplementedError("This backend doesn't support absolute paths.")
NotImplementedError: This backend doesn't support absolute paths.
```

smcoll:
_Still looking this over (want to make sure the default storage behavior isn't broken)_...

Support storage backends which don't have path access by using SpooledTemporaryFile.  In the case of the default storage class, this interim step is probably an inefficiency.  But i confirmed it works for an S3 backend.  Thoughts?

fixes #208

### DIFF
--- a/silk/collector.py
+++ b/silk/collector.py
@@ -1,9 +1,12 @@
+from tempfile import SpooledTemporaryFile
 from threading import local
 
 import cProfile
 import pstats
 import logging
+import marshal
 
+from django.core.files import File
 from django.utils.six import StringIO, with_metaclass
 
 from silk import models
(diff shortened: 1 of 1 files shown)


smcoll:
We're finding that this is pretty hard on resources.

mbeacom:
@smcoll Would you like this revisited? If so, can you please resolve the test failures?

smcoll:
We're no longer using Silk, thanks for asking.

mbeacom:
Closing this issue.  If needed, feel free to re-open this PR and/or pickup the leftover code and start a new branch.

Thanks!

PR #210, opened 2017-08-15
Changed the theme of gprof2dot output to be more inline with rest of silk design

danielbradburn:
I used the green and red colours and FiraSans font so that the graph visualisation fits in a bit better with the rest of silk.


avelis:
@danielbradburn I appreciate you making a theme conscious change to the graphs.

avelis:
Somehow the MySQL configuration is failing authentication in TravisCI will have to investigate later on it.

Issue #219, opened 2017-10-12
Silk is incompatible with django-fullclean


nias-r:
When using [`django-fullclean`](https://github.com/fish-ball/django-fullclean), Silk causes a `ValidationError`: `{'prof_file': ['This field cannot be blank.']}`. 

I believe the cause of the issue is that the model field `prof_file` in the `Request` model has `null=True` but not `blank=True`. From the [Django documentation](https://docs.djangoproject.com/en/1.11/ref/models/fields/#null): 

> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a CharField has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

As `prof_file` is a `FileField` which is string-based, and it is not required to be unique, this advice holds and the field should have `blank=True` and not `null=True`.

p.s. I love this package!

nias-r:
Closing this as it was fixed by #220

PR #220, opened 2017-10-12
Allow prof_file to be blank, not null

nias-r:
See #219


avelis:
@nias-r I realized this after the fact. Can you fix the failing test in TravisCI. Thanks.

PR #221, opened 2017-10-12
gprof2dot had a breaking change in 2017.09.19

nias-r:
gprof2dot introduced a breaking change in the latest version (specifically in [this commit](https://github.com/jrfonseca/gprof2dot/commit/45342280c636dffca74cbdc3e7ac1b4438f57c1c)). As a result the test suite for silk fails.

I've pinned the version to `<2017.09.19`.


avelis:
@nias-r Thanks for catching this and adding it here.

PR #225, opened 2017-11-16
Filter out views that took no time in the database for the most time …

### LINKED ISSUE #217: The dashboard shows views with no queries as most time taken in database
Database: Postgresql.

It seems the views without queries take the most time because it's sorting on a nullable field/aggregation which in postgresql come first when sorting decending.

I don't think django orm supports the "NULLS FIRST" / "NULLS LAST" options, so the nulls should probably be filtered as they are irrelevant for the specific part of the dashboard.

hvdklauw:
…spend in database requests on the dashboard.

Fixes #217

### DIFF
--- a/silk/views/summary.py
+++ b/silk/views/summary.py
@@ -39,7 +39,7 @@ class SummaryView(View):
         return requests
 
     def _time_spent_in_db_by_view(self, filters):
-        values_list = models.Request.objects.filter(*filters).values_list('view_name').annotate(t=Sum('queries__time_taken')).order_by('-t')[:5]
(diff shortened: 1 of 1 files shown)


avelis:
@albertyw Do you see any potential issues here?

albertyw:
I think this is fine, though a test case would be nice.  I'm wondering what requests actually take 0 time; is it a bug that should be fixed?

hvdklauw:
@albertyw No, this is just a nice trick to filter out the views that have no (NULL) time on the database. Because otherwise they show up at the front because by default NULL are ordered first.

hvdklauw:
The same trick as actually already used over here: https://github.com/jazzband/silk/blob/master/silk/views/requests.py#L38

PR #230, opened 2017-12-04
Fix github silk links to point to jazzband

albertyw:
This should be the last few links that pointed to the old mtford90 repository.

Helps with finishing up https://github.com/jazzband/roadies/issues/67


avelis:
@albertyw Thanks for updating the documentation!

PR #233, opened 2017-12-05
Test django 2 in travis

albertyw:
This adds Django 2 to the list of django versions that travis will test against.  

This still requires some work to fix various problems.  Also depends on #232 and #231.


avelis:
@albertyw I am seeing a couple of issues with TravisCI. It appears it can't seem to download Django 2. The other issue is that the projects url settings don't seem to pass certain versions of Django during the testing matrix. Might have to do conditional url patterns or drop the version that is not ok with the pattern change.

avelis:
@albertyw I also remember reading somewhere you wanted to cut a release. The version string is in `setup.py` If you want to break that out into a different mechanic that's fine. FYI It is still currently at `1.0.0`.

avelis:
@albertyw Thanks again for working on this and getting the compatibility for Django 2 in this library!

PR #234, opened 2017-12-08
Fix missing db_time field

albertyw:
Not sure how this got past tests for the original pull request #225, but since the filter is on an annotation, the filter field must match the annotation field.


hvdklauw:
Oh wow, I'm a bit baffled too

PR #235, opened 2017-12-10
Return immediately

Stranger6667:
Trivial changes, without extra variables

### DIFF
--- a/silk/code_generation/django_test_client.py
+++ b/silk/code_generation/django_test_client.py
@@ -20,8 +20,7 @@ def _encode_query_params(query_params):
         query_params = urlencode(query_params)
     except TypeError:
         pass
-    query_params = '?' + query_params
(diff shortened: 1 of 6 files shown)


albertyw:
I personally prefer the style of only returning singular variables instead of expressions.  Returning a singular variable makes functions easier to debug because one can insert a single print/log line without splitting the return and the expression that generates the returned value.  It also more explicitly declares the meaning of the value returned rather than having to guess the value only based on the return expression and function name.  

The change of `selfs` -> `self` is good.

avelis:
@Stranger6667 I am fine with the consolidation effort here to reduce trivial execution lines. It does come at the cost of some readability but I think it's minor.

avelis:
@Stranger6667 Thanks for your contribution to the library. Much appreciated!

PR #245, opened 2017-12-28
fix migration for request.prof_file field

### LINKED ISSUE #244: migrations/0005_increase_request_prof_file_length.py does not match code
The migration make the field nullable while the code uses blank=True. If you run makemigrations the following migrations is generated:

```
# -*- coding: utf-8 -*-
# Generated by Django 1.11.7 on 2017-12-28 13:59
from __future__ import unicode_literals

from django.db import migrations, models
import silk.storage


class Migration(migrations.Migration):

    dependencies = [
        ('silk', '0005_increase_request_prof_file_length'),
    ]

    operations = [
        migrations.AlterField(
            model_name='request',
            name='prof_file',
            field=models.FileField(blank=True, max_length=300, storage=silk.storage.ProfilerResultStorage(), upload_to=b''),
        ),
    ]
```

dennybiasiolli:
Fixes #244


albertyw:
Just to confirm, there were no other migration files generated?

dennybiasiolli:
@albertyw no, this was the only migration file generated by `python manage.py makemigration`

xrmx:
@dennybiasiolli thank you sir!

avelis:
@dennybiasiolli Thanks for making the contribution. Before I approve, would you happen to know if the field will evaluate as an indexable field? I ask because MySQL has a limitation of 191 and we had issues in the past with indexable fields going past max_length of 191. I don't think that is the case here but I wanted to be double sure.

dennybiasiolli:
@avelis I don't think `prof_file` will evaluate as indexable field, because in `models.py` it's defined as

```
prof_file = FileField(max_length=300, blank=True, storage=silk_storage)
```

In `Class Request` the only indexed fields are the following: `path`, `start_time` and `view_name`.

`start_time` is a `DateTimeField`, whilst `path` and `view_name` have already a `max_length=190`, so I think there are no problems with MySQL in this case! 😉

PR #249, opened 2017-12-29
Remove django support 1.7

### LINKED ISSUE #247: Remove support for django 1.7
I think we should stop supporting django 1.7 since it was EOLed 2 years ago (December 2015).  https://www.djangoproject.com/download/#supported-versions

albertyw:
Fixes #247 
Depends on #248


avelis:
@albertyw I don't mind merging this one as well. Looks like it trims a lot of cross compatibility code. Do you mind resolving the conflict here. I don't mind but I don't want to accidentally undo what you wrote in this PR. Although it seems pretty straight-forward.

avelis:
@albertyw Thanks again for refactoring the codebase. Much appreciated.

PR #251, opened 2018-01-04
Use force_text in ResponseModelFactory to avoid b' prefix in django 2

aadu:
…data with b'


avelis:
@aadu Thanks for adding this to the library!

PR #254, opened 2018-01-09
function declaration fix

Yolley:
Function should be declared in <script> tag


avelis:
@Yolley Great catch. I somehow glossed over that and do apologize for that. Thank you for fixing it either way!

PR #266, opened 2018-01-28
 Drop support for Django < 1.11 and remove workarounds

jdufresne:
Per Django's recommendation, as django-silk supports 2.0, it should drop support for Django < 1.11 to ease maintenance and compatibility across Django versions. So long as no deprecation warnings are produced, django-silk should remain forward compatible with the next Django
version.

For more details, see the Django docs:

https://docs.djangoproject.com/en/2.0/releases/2.0/#third-party-library-support-for-older-version-of-django

> Third-party library support for older version of Django
>
> Following the release of Django 2.0, we suggest that third-party app authors drop support for all versions of Django prior to 1.11. At that time, you should be able to run your package’s tests using python -Wd so that deprecation warnings do appear. After making the deprecation warning fixes, your app should be compatible with Django 2.0.

Those wishing to continue using an older Django can pin the version django-silk.

Should significantly reduce testing resources as the test matrix is much smaller.


avelis:
@jdufresne Thank you for cleaning up this library and improving django-silks compatibility for Django 2.0 onward.

PR #270, opened 2018-02-11
Truncate tables when running silk_clear_request_log

### LINKED ISSUE #239: silk_clear_request_log taking longer than 30 minutes
Just installed Silk and ran through all the settings. Everything "works" perfectly (I can see the profiles and queries of the method I set the silk profile on just fine), but the queries seems extremely high for the 2 Django queries I'm making inside the method. I'll be creating an issue after this one to address that issue.

The main issue, however, is that when I tried to clear the logs to see if that's the problem, silk_clear_request_log takes an enormous amount of time. I let it sit for 30 minutes and then went and cooked dinner, came back and it had finally finished. Since I had just installed Silk and run maybe 2 requests, I thought this seemed a bit excessive. Unsure on a fix for this, except perhaps showing some kind of feedback like a progression bar so the user doesn't think the command failed?

albertyw:
Fixes #239 

This should make `./manage.py silk_clear_request_log` run a lot faster by using `TRUNCATE` for non-sqlite databases.  I believe that `objects.all().delete()` won't run `TRUNCATE` on these tables because of foreign key checks.

### DIFF
--- a/silk/management/commands/silk_clear_request_log.py
+++ b/silk/management/commands/silk_clear_request_log.py
@@ -11,13 +11,23 @@ class Command(BaseCommand):
     @staticmethod
     def delete_model(model):
         engine = settings.DATABASES['default']['ENGINE']
-        if 'sqlite' not in engine:
+        table = model._meta.db_table
+        if 'mysql' in engine or 'postgresql' in engine:
             # Use "TRUNCATE" on the table
             cursor = connection.cursor()
-            cursor.execute("TRUNCATE TABLE %s", [model._meta.db_table])
+            if 'mysql' in engine:
+                cursor.execute("SET FOREIGN_KEY_CHECKS=0;")
+            if 'postgres' in engine:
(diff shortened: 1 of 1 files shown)


AuHau:
Just small comment regarding the FKs.

In Postgres you can truncate multiple tables with one command and if you specify all tables which have the FK dependencies just between each other (eq. all Silk tables), then you don't have to turn off the FK checks ;-) 

I have been using this for some time: 
```TRUNCATE TABLE silk_profile, silk_response, silk_profile_queries, silk_request, silk_sqlquery;```

I checked Postgres documentation and multiple-table truncation is possible since 8.1 version, so it should be safe ;-)

Unfortunately this is not possible in MySQL :-/

AuHau:
Hmmm I just saw how the whole command is implemented and without bigger rewrite the previously mentioned technique is not really usable :-/ Damn... Well the it is just FYI :-D

albertyw:
Merging this for now.  This should probably be refactored and tested later.

PR #276, opened 2018-04-09
Revert "Opening sql queries in new tab is very useful"

### LINKED ISSUE #271: Force opening a new window for SQL queries is very annoying
PR #253 made a change such that the SQL link forces a new window to be opened. This can be very frustrating for some workflows. Every browser has a way to easily open a link in a new tab/window, forcing this on a user is very annoying as there is no easy way to prevent the link being opened in a new tab/window.

Are there many people in favour of reverting this? If so, leave a :+1:

albertyw:
This reverts commit 6b0b22e48dd772e46d23e68efb95cc08935270f5.

Fixes #271

### DIFF
--- a/silk/templates/silk/sql.html
+++ b/silk/templates/silk/sql.html
@@ -8,14 +8,7 @@
 {% block js %}
   <script type="text/javascript" src="{% static 'silk/lib/sortable.js' %}"></script>
   {{ block.super }}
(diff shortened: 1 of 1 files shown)


avelis:
I don't mind merging this in. Maybe in the future we make this a setting that can be set True or False.

PR #279, opened 2018-04-25
Fix index error

drppi44:
Sometimes this line raise indexerror
i didn't research this case

PR #280, opened 2018-04-30
Fix garbage collection logic for small tables

albertyw:
This makes sure that there will be no `IndexErrors` when calling the `garbage_collect` method.  I'm guessing this may be possible if there are fewer `Request`s than the target number of requests.

cc @avelis @drppi44 

Replaces #279

### DIFF
--- a/silk/models.py
+++ b/silk/models.py
@@ -148,6 +148,7 @@ class Request(models.Model):
         requests = cls.objects.order_by('-start_time')
         if not requests:
             return
+        target_count = max(len(requests)-1, target_count)
         time_cutoff = requests[target_count].start_time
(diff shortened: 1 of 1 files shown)


silk/models.py
```diff
@@ -148,6 +148,8 @@ def garbage_collect(cls, force=False):
         requests = cls.objects.order_by('-start_time')
         if not requests:
             return
+        if len(requests)-1 < target_count:
```
avelis: Couldn't we combine this line with the line above on 149? I know it lengthens the if check with an or but ensures only one return statement.

albertyw:
@avelis I made that update.

avelis:
@albertyw Thanks!

drppi44:
thatnks!

PR #281, opened 2018-05-04
Added a distribution tab to visualize request time distributions over time

danielbradburn:
This PR adds a distribution tab (at the moment opt in using the SILK_DISTRIBUTION_TAB configuration setting). We are using this at crunchr for continous benchmarking to ensure we don't introduce changes to our codebase that degrade the performance of our backend. This might be particularly useful for benchmarking django apps over time or versions.


requirements.txt
```diff
@@ -11,3 +11,4 @@ Django>=1.11
 freezegun>=0.3
 factory-boy>=2.8.1
 gprof2dot>=2016.10.13,<2017.09.19
+dealer>=2.0.5
```
albertyw: Checking out [dealer](https://github.com/klen/dealer), it seems dealer may be out of date.  Have you tried testing with django >= 1.10 (https://github.com/klen/dealer/issues/24)?

silk/views/filterable_requests_view.py
```diff
@@ -0,0 +1,102 @@
+try:
+    # Django>=1.8
```
albertyw: As of fa1f5425dfd811d200f20856ba195fb8062bc491, django-silk does not support django<1.11

silk/static/silk/lib/url.js
```diff
@@ -0,0 +1,199 @@
+(function() {
```
albertyw: Is this an unminified copy of `url.min.js`?  Do we need both?

silk/templates/silk/profiling.html
```diff
@@ -94,13 +94,13 @@ <h4>Profile</h4>
                    value="OverallTimeFilter"
                    name="filter-overalltime-typ"/>
             <input type="text"
-                   placeholder="seconds"
+                   placeholder="milliseconds"
```
albertyw: Is this a bugfix?

silk/config.py
```diff
@@ -2,15 +2,37 @@
...
+        #
+        #  - it only works for git
+        #  - it's using the protected _repo attribute from dealer
+        #  - the timestamp becomes visible in the revision
+        #
+        # However it is better than order commit hashes alphabetically
+        timestamp = auto._repo.git('show -s --format=%at ' + revision)
+        revision = ' # '.join([timestamp.decode(), revision])
+    except:
+        pass
```
albertyw: Does this make more sense to be `revision = ''`?

And does the rest of the app work correctly with an empty string or `None` value?  There seem to be several areas of the app that filter or group by revision value.

requirements.txt
danielbradburn: I will take a look for an alternative to ``dealer``. I think the revision fuctionality is not that useful in production, at least we don't use it, we set ``SILKY_REVISION`` to a version number in ``settings.py``. But it is quite useful for benchmarking locally, for example for comparing a number of commits. If I can't find a better alternative I guess another option is to just remove this dependency and let users set to the revision number using whichever method they prefer (probably nice to have an example in the readme)

silk/static/silk/lib/url.js
danielbradburn: Yes and no, I will remove it.

silk/views/filterable_requests_view.py
danielbradburn: Ah good, improves this code!

requirements.txt
danielbradburn: I will check out the failing tests.

silk/templates/silk/profiling.html
danielbradburn: I guess this is a typo fix, since the placeholder says 'seconds' but the filter actually accepts 'milliseconds', do you think it makes sense to make a separate PR for this?

silk/config.py
moagstar: @albertyw I removed the dealer dependency now - I think it's simpler just to let users to set the revision themselves using whichever method they prefer. This is what we have ended up doing when running this branch in production.

silk/static/silk/lib/url.js
moagstar: @albertyw I removed this unnecessary file

moagstar:
@albertyw @avelis Apologies for letting this PR sit for so long. I have fixed the tests, and am curious if there is any further feedback or if this can be merged. Let me know if you have any questions.

silk/migrations/0008_request_revision.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
auvipy, line 1: please remove this migrations and regen again with django 2.2+ only.

silk/request_filters.py
```diff
@@ -206,16 +206,21 @@ def __init__(self, value):
         super(MethodFilter, self).__init__(value, method=value)
 
 
-def filters_from_request(request):
+class RevisionFilter(BaseFilter):
+    def __init__(self, value):
+        super(RevisionFilter, self).__init__(value, revision=value)
```
auvipy, line 211: no need of expliclit RevisionFilter, self) in super()

silk/templates/silk/inclusion/request_filter.html
```diff
@@ -0,0 +1,174 @@
...
+                    <option value="" {% if not filters.method.value %}selected{% endif %}></option>
+                    {% for method in options_methods %}
+                        <option {% if method == filters.method.value %}selected{% endif %}>{{ method }}</option>
+                    {% endfor %}
+                </select>
+            </label>
+        </div>
+
+
+    </div>
```
auvipy, line 174: no newline

silk/urls.py
```diff
@@ -89,4 +91,19 @@
         view=CProfileView.as_view(),
         name='cprofile',
     ),
+    url(
```
auvipy, line 94: using path() would be better choice as url() is deprected

silk/views/distribution.py
```diff
@@ -0,0 +1,53 @@
+import csv
+import contextlib
+from six import StringIO
```
auvipy, line 3: six is not needed

auvipy (changes requested):

Archmonger:
@moagstar do you have time to address @auvipy 's change requests?

moagstar:
> @moagstar do you have time to address @auvipy 's change requests?

I am no longer working at crunchr, so I don't think have permissions to push to this branch any longer (in fact I made this PR with my crunchr account). I would probably have to create a new PR from my own fork. I am not sure when I would have time to look at this though, since we don't use silk at my new job and personal time for programming is quite limited at the moment.

auvipy:
> > @moagstar do you have time to address @auvipy 's change requests?
> 
> I am no longer working at crunchr, so I don't think have permissions to push to this branch any longer (in fact I made this PR with my crunchr account). I would probably have to create a new PR from my own fork. I am not sure when I would have time to look at this though, since we don't use silk at my new job and personal time for programming is quite limited at the moment.

thanks for your work, i might end up opening a cleaner version of your work

eduzen:
Hey @auvipy did you have time to check it? If not, I would like to help!

PR #284, opened 2018-05-14
Allow disabling all sql logging with '*' in SILKY_IGNORE_QUERIES

naytec:
I don't want/need to log sql queries.
Currently can't disable it for all requests, so how about doing it this way?
One can add set SILKY_INGORE_QUERIES to `['*']` or just `*`.

Does it make sense?

### DIFF
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -49,6 +49,12 @@ def _should_intercept(request):
     return not (silky or ignored)
 
 
+# whether sql logging should be enabled at all
+_should_log_queries = not (
+    config.SILKY_IGNORE_QUERIES and config.SILKY_IGNORE_QUERIES[0] == "*"
+)
+
+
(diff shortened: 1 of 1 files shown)


albertyw:
Looking at the current logic in sql.py, it seems you could ignore all sql queries by setting `SILKY_IGNORE_QUERIES = ['']`?

naytec:
@albertyw Ah right, indeed. I haven't noticed that!
In this case however, there's no need to alter SQLCompiler.execute_sql in middleware. But that probably doesn't make a big difference anyway.

Thanks!

avelis:
@albertyw Is this PR ok to close without merging? Should we add some documentation on how to achieve @naytec's request?

albertyw:
Yeah, I think we should close this without merging.  Updating documentation would be nice, but we should also fix #286

PR #295, opened 2018-07-12
Added a simple view in request detail context allowing to get python profile

laurentb2:
After checking multiple times, I figured out that python profile is provided only on the profile detailed page which is reachable from the profiling view only if some dynamic, decorator or context manager profile are defined. I simply created a new view allowing to get the python profile for each and every queries.

capture


avelis:
@laurentb2 Thanks for the contribution.

PR #296, opened 2018-07-20
Modernize middleware

gforcada:
As django-silk no longer supports django < 1.10 the MiddlewareMixin is
no longer needed.

### DIFF
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -5,7 +5,6 @@ from django.db import transaction, DatabaseError
 from django.urls import reverse, NoReverseMatch
 from django.db.models.sql.compiler import SQLCompiler
 from django.utils import timezone
-from django.utils.deprecation import MiddlewareMixin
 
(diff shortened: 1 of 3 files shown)


silk/middleware.py
```diff
@@ -57,7 +56,19 @@ def process_request(self, request):
         return
 
 
-class SilkyMiddleware(MiddlewareMixin):
+class SilkyMiddleware(object):
```
albertyw: This shouldn't need to inherit from object anymore since we are now python 3.

gforcada: Thanks for the review, indeed, that is no longer needed.

Unfortunately I'm going on vacations and I can not fix that myself, I could gladly do that as soon as I'm back around August 15th

gforcada:
@albertyw I fixed your review. Would you mind getting another look? TIA!

gforcada:
@avelis thanks!

Issue #290, opened 2018-06-11
silk_clear_request_log fails on Postgres


devmonkey22:
django-silk==3.0.0

The SQL executed to clear the request logs fail with multiple errors on Postgres 9.5.12.

The errors include:

- django.db.utils.ProgrammingError: syntax error at or near "'silk_profile'" LINE 1: ALTER TABLE 'silk_profile' DISABLE TRIGGER ALL;
    -  Postgres needs unquoted or double quotes, but was given single quotes.
- django.db.utils.NotSupportedError: cannot truncate a table referenced in a foreign key constraint DETAIL:  Table "silk_profile_queries" references "silk_profile". HINT:  Truncate table "silk_profile_queries" at the same time, or use TRUNCATE ... CASCADE.

These seem to stem from changes for PR #270. 

```
Traceback (most recent call last):
  File "oc/manage.py", line 18, in <module>
    main()
  File "oc/manage.py", line 15, in main
    execute_from_command_line(sys.argv)
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/silk/management/commands/silk_clear_request_log.py", line 41, in handle
    Command.delete_model(silk.models.Profile)
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/silk/management/commands/silk_clear_request_log.py", line 21, in delete_model
    cursor.execute("ALTER TABLE %s DISABLE TRIGGER ALL;", [table])
  File "/home/user/dev/oc/ENV/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
[...]

PR #299, opened 2018-08-15
Disable postgres USER triggers

gforcada:
To be able to disable ALL triggers on a database, one needs a user that can disable system triggers.

Should fix https://github.com/jazzband/django-silk/issues/290


avelis:
@gforcada I am not a PG user here. So in theory, I am ok with the change.

devmonkey22:
Thanks @gforcada, the patch from `ALL` to `USER` works for me too in PG.

avelis:
@gforcada Let me see what I can do. I appreciate the patience and all the contributions this library gets.

mOscuro:
Hi! any news about a possible release? I'm getting impacted by this as well. Thanks!

Issue #319, opened 2018-12-11
Ignored request triggering raise_middleware_error


rwlogel:
Our project is using `django-health-check` and whenever the health check endpoint it requested it triggers the `middleware_error`.  Adding the health check endpoint to the `SILKY_IGNORE_PATHS` didn't work as expected.

Upon startup if only ignored paths are requested everything works fine.  As soon as a path which should be evaluated by SILK is requested then all subsequent ignored paths trigger some of the processing, even when they are supposed to be ignored.  This causes the incompatibility with `django-health-check` to trigger the `middleware_error`.

If a path is supposed to be ignored by SILK then it shouldn't be wrapping the SQL queries or checking if the middleware is configured correctly.

PR #320, opened 2018-12-11
If there is no DataCollector().request then don't wrap sql queries

rwlogel:
This is a potential fix for issue #319


rwlogel:
I updated the unit tests to account for the changes.

PR #335, opened 2019-03-06
Change fields from CharField with max length to be TextFields that have no max length

KevinGrahamFoster:
Silk produced this error when I access a URL longer than 190 chars in my app:

```
django.db.utils.DataError: value too long for type character varying(190)
```

Looking at the code I realized there are a few fields that are CharFields with a max length which I could not see a reasonable reason to enforce a max length on them.  Yes I realize a 190 char url is unwieldy to type, but I don't see why this limit should be enforced.  As I understand them TextFields are someone similar in performace to CharFields so I changed all those fields that use them.

Thanks, if I am off on something here, looking forward to hearing about it.


silk/models.py
```diff
@@ -54,14 +54,14 @@ def __init__(self, d):
 
 class Request(models.Model):
     id = CharField(max_length=36, default=uuid4, primary_key=True)
-    path = CharField(max_length=190, db_index=True)
+    path = TextField(db_index=True)
```
mbeacom: Eh - I'd be more inclined to merge this if we stuck with `CharField` with a max of `255` - Were you able to test `255` for `path` and `view_name` ? - `method` could probably remain the same.

mbeacom:
Closing due to unresponsiveness - Feel free to re-open this if you'd like to revisit it.

PR #356, opened 2019-07-30
Add docker setup

Chris7:
This adds a docker setup to the project. It's easier for me to just setup a dockerfile than to get my host setup right :)


avelis:
@Chris7 Does this add docker as a requirement to development/testing with the project? If so I would push to update the docs to say so.

Chris7:
It doesn't add any dependencies. I'll update the docs for those who like using docker. I wanted to get a PR up to see if it was even something you'd be interested in. Personally, it's great because I no longer have to care about virtual environments/etc.

PR #372, opened 2019-11-03
Allow custom request and response model factories

yunier-rojas:
Sometimes, it's needed custom factories for request and
response models to handle incoming request in the
middleware. You could solve this by extending the
middleware but probably end up copying a lot of code.
With this development, you can configure custom classes
to solve this.

`SILKY_REQUEST_MODEL_FACTORY_CLASS` and
`SILKY_RESPONSE_MODEL_FACTORY_CLASS` are settings that
allow customization of factories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,4 @@ fabric.properties
 
 # Virtual env
 .venv*
+/venv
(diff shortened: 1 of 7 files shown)


mbeacom:
@yunier-rojas Travis seems to be failing on python 2.7. 🤔 
Honestly, I wouldn't mind `django-silk` dropping 2.x support since it's EOL as of January anyway, but for now, we'll probably want to ensure tests are passing across the board until that is removed.

auvipy:
I will do a final review

auvipy (changes requested):
can you re base?

PR #379, opened 2019-12-06
Fix hiding of sensitive data in requests

RaphaelKimmig:
The previous implementation does not correctly handle:

- prefixes in json
- suffixes in forms & json
- case sensitivity in forms & json
- nested values in json

These changes should fix the above issues.


nasirhjafri:
> And make the Travis green

Tests are only failing on `Python 2.7`. Once rebased, hopefully, Travis will be green too.

RaphaelKimmig:
Rebased, let's hope that the tests will be nice and green :-)

PR #387, opened 2020-01-04
traceback only when needed

ilndinesh:
calculate traceback only if we need to wrap sql


ilndinesh:
looks like not update to tests are needed for this

auvipy:
can you check here https://github.com/jazzband/django-silk/blob/master/project/tests/test_view_sql_detail.py and related files?

Archmonger:
@ilndinesh Can you rebase this to the latest django-silk:master

SebCorbin:
This can be considered as code cleanup/performance optimization and could be merged as is (provided the tests pass)

Archmonger (approved):

PR #393, opened 2020-01-21
Refactor QA setup

aleksihakli:
- Utilize `tox` with `tox-travis` for more UX friendly test orchestration
- Utilize `pytest` with `pytest-django` and `pytest-cov` for more readable test output
- Utilize `setuptools_scm` for inferring package version
- Improve test matrix definitions for database and Python version combinations
- Skip false-positive test for cProfiling output


.travis.yml
```diff
@@ -1,48 +1,57 @@
...
 dist: xenial
-sudo: false
+language: python
+cache: pip
 python:
-- '3.5'
-- '3.6'
-- '3.7'
-- '3.8'
+  - 3.6
```
nasirhjafri: Any specific reason why have we removed python 3.5 support?

.travis.yml
```diff
@@ -1,48 +1,57 @@
...
-matrix:
-  exclude:
-  - python: '3.5'
-    env: DJANGO="Django>=3.0"            DB=postgresql DB_NAME=travis_ci_test
-  - python: '3.5'
-    env: DJANGO="Django>=3.0"            DB=sqlite3 DB_NAME=db.sqlite3
-  - python: '3.5'
-    env: DJANGO="Django>=3.0"            DB=mysql DB_NAME=mysql_db
-  fast_finish: true
+  - DJANGO=1.11     DATABASE=sqlite3
```
nasirhjafri: We dropped Django 1.11 support in last release.

aleksihakli: OK, I just added that to the test matrix to see if the tests pass here. Django 1.11 is a LTS release which will still be supported by DSF until April 2020. 

I have no argument over support decisions, I'm just trying to offer improvements for the infrastructure. You decide what looks good to you!

.travis.yml
aleksihakli: @nasirhjafri this is is just a practice I have adopted from some other packages I maintain.

Python 3.5 is not supported by newer Django versions and it simplifies things to offer support on 3.6+ feature wise. 3.6 has a lot of forwards compatible features. Python 3.5 is also dropping in usage numbers as most distributions offer 3.6+ as the stock distribution.

nasirhjafri: Django 2.2  is still supporting python 3.5 since we are supporting Django 2.2 let's also support Python 3.5 too.

.travis.yml
nasirhjafri: Okay, cool, let's keep Django 1.11 for now. We will drop it later in April.

aleksihakli: @nasirhjafri if the version is supported by the codebase and there is no sudden need to remove support I'd opt to support LTS releases as long as they are being maintained by DSF as that will make a lot of people running older code bases happy :)

.travis.yml
aleksihakli: OK, I'll add it to the matrix!

tox.ini
```diff
@@ -0,0 +1,45 @@
+[travis]
+python =
+    3.6: py36
+    3.7: py37
+    3.8: py38
+    pypy3: pypy3
+
+[travis:env]
+DJANGO =
+    1.11: django111
```
auvipy: not needed

tox.ini
```diff
@@ -0,0 +1,45 @@
...
+    py{35}-django{111,22}-{sqlite3,mysql,postgresql}
+    py{36,37,38,py3}-django{111,22,30,master}-{sqlite3,mysql,postgresql}
+
+[testenv]
+changedir = {toxinidir}/project
+deps =
+    -r project/requirements.txt
+    mysql: mysqlclient
+    postgresql: psycopg2
+    django111: django>=1.11,<1.12
```
auvipy: not needed

tox.ini
nasirhjafri: Django 1.11 is an LTS release that will still be supported by DSF until April 2020. So we'll be providing support for 1.11 for now.

aleksihakli: Is there an actual downside for supporting Django 1.11 LTS for 5 more months? The test matrix passes for it and everything should work smoothly as it is.

aleksihakli:
@nasirhjafri @auvipy @jezdez the `django-silk` QA improvements are ready for upstreaming. There is one false-positive case in the tests in the cProfile section that results from a small difference in timings with Python 3.6 interpreter, but otherwise all tests pass and the test bench is considerably improved.

The test also fails on PyPy3 in addition to Python 3.6 and should be refactored to account for differing cProfile outputs or disabled altogether if it is not necessary:

image

tox.ini
aleksihakli: Well since the support was dropped in 4.0.0 I've removed it from the matrix as well, April will be soon and no point in reintroducing support after a major release.

tox.ini
aleksihakli: Roger.

setup.py
```diff
@@ -1,21 +1,18 @@
...
 
 # allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
-README = open('README.md', 'rb').read().decode("UTF-8")
-
 setup(
     name='django-silk',
+    use_scm_version=True,
     version='4.0.0',
```
jezdez: The version parameter needs to go.

docs/conf.py
```diff
@@ -51,9 +47,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.3'
+version = pkg_resources.get_distribution("django-silk").version
 # The full version, including alpha/beta/rc tags.
-release = '0.3'
+release = version
```
jezdez: Let's use the way release and version is distinguished in the setuptools-scm docs; https://github.com/pypa/setuptools_scm/#usage-from-sphinx

project/tests/test_profile_parser.py
```diff
@@ -1,22 +1,25 @@
...
+
 from silk.utils.profile_parser import parse_profile
 
 
 class ProfileParserTestCase(TestCase):
-
+    @unittest.skipIf(
+        os.environ.get("CONTINUOUS_INTEGRATION") == "true",
+        "Skip profiling integration test when running CI on due to frequent false negatives"
+    )
```
jezdez: Hmm, the test should be written differently then.

setup.py
aleksihakli: Fixed.

docs/conf.py
aleksihakli: Fixed.

project/tests/test_profile_parser.py
aleksihakli: I very much agree: however, the test is not simple to fix as the result is a more complex object and ideally the format should be validated, not the contents itself.

Hence, the test would ideally use some sort of a regex matcher for a matching list that is inside the list of profiling results.

This test and the max body size test are also the ones that fail on PyPy3 at the moment.

I would opt a skip in this PR and then a fix in a later PR that updates the test to a more appropriate format, but I am not interested in refactoring the tests at this time.

aleksihakli:
@jezdez I believe the version is ready for merging now. Thanks for the accurate review insights!

aleksihakli:
Seems like it's the same test. For whatever reason it isn't skipping the test with the `@unittest.skipIf` decorator. I think I'll refactor the test after all to validate the format of data instead of it's exact output so that the test supports PyPy as well.

aleksihakli:
Test refactored to test output format instead of specific tool output, seems to be working. Upstreaming the changeset.

Issue #410, opened 2020-03-31
_mask_credentials uses UGC in a regex substitution


barm:
I recently ran into an issue when testing passwords with whitespace characters on an app with silk installed. I found that if I have multiple consecutive spaces in a password (which silk treats as a sensitive field and tries to mask), the regex substitution in `_mask_credentials` fails. This is because the user input is used as a pattern in a `re.sub` and so if the user uses certain characters, they are interpreted as a regex pattern. Space here in the `raw_body` is a literal `+`. This seems unsafe. 

```python
for res in results:
    body = re.sub(res[1], RequestModelFactory.CLEANSED_SUBSTITUTE, body)
```
From: https://github.com/jazzband/django-silk/blob/master/silk/model_factory.py#L131

Is there any reason we can't do a simple string replace instead?
```python
for res in results:
    body = body.replace(res[1], RequestModelFactory.CLEANSED_SUBSTITUTE)
```

PR #413, opened 2020-04-07
Replace the call to re.findall with re.sub in _mask_credentials so matched values are not treated as regex patterns

ThePumpingLemma:
This fixes https://github.com/jazzband/django-silk/issues/410

This also fixes a DoS attack:
* Form is submitted with multiple empty fields that match the mask filter
* Substitution loop repeatedly calls `re.sub` with an empty pattern
* Cleansed string is inserted between every pair of characters
* Loop iterations causes exponential growth of body string, preventing the request from ever finishing

### DIFF
--- a/project/tests/test_sensitive_data_in_request.py
+++ b/project/tests/test_sensitive_data_in_request.py
@@ -36,6 +36,9 @@ class MaskCredentialsInFormsTest(TestCase):
     def test_mask_credentials_handles_suffixes(self):
         self.assertNotIn("secret", self._mask("username-with-suffix=secret"))
 
+    def test_mask_credentials_handles_regex_characters(self):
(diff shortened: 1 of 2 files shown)


project/tests/test_sensitive_data_in_request.py
```diff
@@ -36,6 +36,9 @@ def test_mask_credentials_handles_prefixes(self):
     def test_mask_credentials_handles_suffixes(self):
         self.assertNotIn("secret", self._mask("username-with-suffix=secret"))
 
+    def test_mask_credentials_handles_regex_characters(self):
+        self.assertNotIn("secret", self._mask("password=secret++"))
```
barm, line 40: 👍

PR #417, opened 2020-04-15
Clear DB error when configuring silk to use a non-' default' database

eshxcmhk:
resolve```django.db.utils.ProgrammingError: relation "silk_profile" does not exist```


nasirhjafri:
@eshxcmhk Could you please update your comments in English? Thanks,

Issue #429, opened 2020-07-21
Requests list should be one dimensional


Kern3l:
Hey!

"Requests" view really begs to be a list, maybe even rows of a table (that can be sorted by just clicking on a header, or even it could be filtered by just clicking on a value of later rows). It would be easier to find what you want visually - moving your eyes in one dimension is quicker than in 2 as it is now.

Thanks in advance!

mierz00:
I agree completely with @Kern3l. I'd love to have a list. 

Would you guys be willing to accept a pull request with this feature? 

I'm imagining a simple selector in the nav bar that allows you to choose a list or classic view. 

@nasirhjafri @albertyw

nasirhjafri:
@mierz00 sure, please create a PR and tag me for review, thanks!

auvipy:
thank you

PR #440, opened 2020-09-12
Add row view for requests page

mierz00:
Adds a list/row view for the requests page #429.

### Changes
- Added a selector to choose between the card view and the row view.
- Added SCSS for the row/list view.
- Corrected the error when trying to run "gulp watch" for the SCSS files. 
- Corrected the README instructions for installing dependencies (requirements.txt doesn't exist in the project root).

My only gripe at the moment is that when we click on a request and then click on the back button, all the user parameters are cleared. It would be nice to keep track of what the user already selected.


nasirhjafri (approved):
LGTM 👍

nasirhjafri:
Thanks, @mierz00

PR #445, opened 2020-10-02
Replace url with path for Django 4.0

MuslimBeibytuly:


auvipy (changes requested):
why not use path()?

MuslimBeibytuly:
> why not use path()?

replaced using path, pls review:)

auvipy (commented):
do we need to explicitly mention route='' , views= in the path() ?
https://docs.djangoproject.com/en/3.1/ref/urls/

though it looks nice

MuslimBeibytuly:
> do we need to explicitly mention route='' , views= in the path() ?
> https://docs.djangoproject.com/en/3.1/ref/urls/
> 
> though it looks nice
it isn't required, but subjectively transparent keyword args seems to be better:)

auvipy (approved):

MuslimBeibytuly:
@auvipy when should I be expecting it merged and released?:)

auvipy:
one question, did you try this approach locally?

MuslimBeibytuly:
> one question, did you try this approach locally?

I've ran tests, if that's what you're asking about:)

PR #458, opened 2021-01-27
Add remote storage support

k4rl85:
Some years ago i had to hack a solution to add support for remote storage.
The years goes on and now i have to update django silk to support last release of Django and in consequence my customization.

I see there is an issue #208 and a proposed solution #209 but it does not seems to have a lot of traction.

I wonder if you are interested to support this functionality and after some refine of my hack i made this pr.

The following changes are introduced:
1. Add support for remote storages by using django storage interface
2. Add the settings `SILKY_PROF_FILE_PREFIX` to allow set a prefix
3. Add request path to the profile file path

The 2nd point was necessary in my implementation to use the same remote storage with different version of the code while the 3rd is just a small improvement to retrieve more easily the profile files of the same request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,6 +488,14 @@ To enable query analysis when supported by the dbms a config var can be set in o
 SILKY_ANALYZE_QUERIES = True
 ```
 
+### Set profile filename prefix
+
+Allow to set a prefix to the profile filename
+
+```python
+SILKY_PROF_FILE_PREFIX = "prof/"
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
can you please add unit tests for the proposed changes?

Archmonger:
@k4rl85 can you rebase this to the latest django-silk:master

PR #481, opened 2021-06-24
Improve performance by splitting query details off into seperate table

ErwinJunge:
This PR aims to improve the performance of silk by splitting the query details off into a seperate database table.

Some background information: The current system stores the details of the queries directly inside the query table itself. This leads to a very wide table and consequently few tuples per page, which makes doing aggregates over large tables very expensive in IO since the database has to load all the pages into memory.

What this PR does to fix that is pull the details that aren't used in the overview pages out of the main table and into a seperate table. This decreases the width of the main table, leading to more tuples per page and therefore less IO when generating overviews.

Specifically, in our production environment the sqlquery table has >3M rows. This lead to the overview page taking more than a minute to load and therefore timeout. With these changes, the overview page takes ~9s. Still long, but at least it doesn't timeout anymore.


ErwinJunge:

silk/migrations/0010_fill_query_details.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.2.24 on 2021-06-24 09:54
+
```
auvipy, line 2: is this migration handwritten? or auto generated

ErwinJunge: The code inside the migration was handwritten, the file was autogenerated using `./manage.py makemigrations silk --empty --name fill_query_details`

Archmonger:
@ErwinJunge This is a candidate for merge if tox tests can be passed.

auvipy (changes requested):
could you please rebase?

PR #482, opened 2021-06-24
Don't return null values when computing "Most Time Overall" summary

ErwinJunge:
This PR aims to fix a bug where this summary could return a list of empty timing values if there are null values in the `time_taken` field.


ErwinJunge:

auvipy (changes requested):
can you check the build failures? also shouldn't we update unit tests for the change?

ErwinJunge:
> can you check the build failures? also shouldn't we update unit tests for the change?

The build failures aren't related to the code changes in the PR. In fact, they're also present in https://github.com/jazzband/django-silk/pull/479/checks which doesn't do any code changes at all. It looks like master is broken.

Regarding updating the unit tests: I agree, that was a bit of an oversight on my end. I'll update them asap.

PR #484, opened 2021-06-24
Ensure sorting between most db queries requests

ErwinJunge:
This PR ensures that the result in the "Most Database Queries" is actually sorted in descending order. The previous code could result in the results being out of order since the ordering was applied seperately per view_name.

### DIFF
--- a/silk/views/summary.py
+++ b/silk/views/summary.py
@@ -51,7 +51,7 @@ class SummaryView(View):
                 requests.append(r)
             except IndexError:
                 pass
-        return requests
+        return sorted(requests, key=lambda item: item.t, reverse=True)
 
     def _create_context(self, request):
         raw_filters = request.session.get(self.filters_key, {})


ErwinJunge:

auvipy:
done

auvipy (changes requested):
can you check the build failures? also shouldn't we update unit tests for the change?

ErwinJunge:
> can you check the build failures? also shouldn't we update unit tests for the change?

The build failures aren't related to the code changes in the PR. In fact, they're also present in https://github.com/jazzband/django-silk/pull/479/checks which doesn't do any code changes at all. It looks like master is broken.

Regarding updating the unit tests: I agree, that was a bit of an oversight on my end. I'll update them asap.

auvipy:
but test needed for all the proposed changes

Andrew-Chen-Wang:
@auvipy ref: https://github.com/psycopg/psycopg2/issues/1293 looking to see if anyone else has a solution in jazz band discussion. If not, then we can merge #486

auvipy (approved):

Issue #489, opened 2021-08-17
ValueError: Unknown options: analyze (Sqlite backend)


Archmonger:
Django Silk `ANALYZE_QUERIES` does not work with Sqlite3 when using Django 3.2.6.

Django silk fetched via `pip install -e git+https://github.com/jazzband/django-silk.git#egg=django-silk`.

Django-silk latest commit: [`7f03cfef0b554af541dc53fc6d5bee850469098a`](https://github.com/jazzband/django-silk/commit/7f03cfef0b554af541dc53fc6d5bee850469098a).

```python
# settings.py
SILKY_AUTHENTICATION = True
SILKY_AUTHORISATION = True 
SILKY_ANALYZE_QUERIES = True # This appears to result in the problem taking place.
SILKY_PYTHON_PROFILER = True
SILKY_PYTHON_PROFILER_BINARY = True
```

This issue is not present on Django 3.1 using Django-silk 4.1

### Traceback
```cmd
Traceback (most recent call last):
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\asgiref\sync.py", line 482, in thread_handler
    raise exc_info[1]
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\django\core\handlers\exception.py", line 38, in inner
    response = await get_response(request)
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\django\core\handlers\base.py", line 233, in _get_response_async
    response = await wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\asgiref\sync.py", line 444, in __call__
    ret = await asyncio.wait_for(future, timeout=None)
  File "C:\Users\username\AppData\Local\Programs\Python\Python39\lib\asyncio\tasks.py", line 442, in wait_for
    return await fut
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\asgiref\current_thread_executor.py", line 22, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\username\Documents\Repositories\Conreq\.venv\lib\site-packages\asgiref\sync.py", line 486, in thread_handler
    return func(*args, **kwargs)
  File "c:\users\username\documents\repositories\conreq\.venv\src\django-silk\silk\profiling\profiler.py", line 158, in wrapped_target
[...]

PR #493, opened 2021-09-18
Add tests that access the actual DB

marius-mather:
The project's existing tests all seem to mock DB access, unless I'm missing something. This PR adds some simple tests that actually access the DB. This allows us to actually catch compatibility errors like those in #489 ,

Let me know what you think.

### DIFF
--- a/project/tests/factories.py
+++ b/project/tests/factories.py
@@ -4,6 +4,8 @@ import factory.fuzzy
 
 from silk.models import Request, Response, SQLQuery
 
+from example_app.models import Blind
+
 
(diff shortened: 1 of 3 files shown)


marius-mather:
Once we have these tests that actually hit the DB, we can get them all passing with a try/except around the call to `explain_query_prefix`: https://github.com/marius-mather/django-silk/pull/2/

marius-mather:
I've added a workaround for this issue. I think this is ready to be tested and reviewed, if someone wants to approve the workflow and have a look.

auvipy (approved):

PR #494, opened 2021-09-21
Make sensitive keys set configurable

jeffrey-effendy:
Currently the masking logic has hard coded set of sensitive keys under RequestModelFactory to mask the body value. There are times where I need to add values into the existing keys. It is hard with the current implementation.

Since the masking logic is already implemented, it would be nice to have the sensitive keys configurable from Django settings. This will make the masking more flexible for many users.

For this I create a new key in config
`SILKY_SENSITIVE_KEYS `

with default value
`{'username', 'api', 'token', 'key', 'secret', 'password', 'signature'}`
so it does not affect any previous users.

Let me know if this is a good addition

### DIFF
--- a/silk/config.py
+++ b/silk/config.py
@@ -30,7 +30,8 @@ class SilkyConfig(metaclass=Singleton):
         'SILKY_STORAGE_CLASS': 'silk.storage.ProfilerResultStorage',
         'SILKY_MIDDLEWARE_CLASS': 'silk.middleware.SilkyMiddleware',
         'SILKY_JSON_ENSURE_ASCII': True,
-        'SILKY_ANALYZE_QUERIES': False
+        'SILKY_ANALYZE_QUERIES': False,
+        'SILKY_SENSITIVE_KEYS': {'username', 'api', 'token', 'key', 'secret', 'password', 'signature'}
     }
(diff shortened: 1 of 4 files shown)


silk/config.py
```diff
@@ -30,7 +30,8 @@ class SilkyConfig(metaclass=Singleton):
         'SILKY_STORAGE_CLASS': 'silk.storage.ProfilerResultStorage',
         'SILKY_MIDDLEWARE_CLASS': 'silk.middleware.SilkyMiddleware',
         'SILKY_JSON_ENSURE_ASCII': True,
-        'SILKY_ANALYZE_QUERIES': False
+        'SILKY_ANALYZE_QUERIES': False,
+        'SILKY_SENSITIVE_KEYS': {'username', 'api', 'token', 'key', 'secret', 'password', 'signature'}
```
auvipy, line 34: this change also need docs for usage

auvipy (changes requested):

silk/config.py
jeffrey-effendy: thanks, added usage docs under README.md for your review

auvipy (approved):

PR #498, opened 2021-09-27
explicit "id" fields to `SQLQuery` and `Profile`

daadu:
This is to avoid unnecessary migration files created inside package by user when running `makemigrations` - becasuse after Django 3.2 "default id field" is configurable project wide by user with `DEFAULT_AUTO_FIELD`.

Also test for checking pending migration added.

### DIFF
--- a/project/tests/test_models.py
+++ b/project/tests/test_models.py
@@ -3,7 +3,8 @@ import datetime
 import uuid
 import pytz
 
-from django.test import TestCase
+from django.core.management import call_command
+from django.test import TestCase, override_settings
 from django.utils import timezone
 
 
@@ -489,6 +490,25 @@ class SQLQueryTest(TestCase):
         self.obj.delete()
 
         self.assertNotIn(self.obj, models.SQLQuery.objects.all())
+
+
+class NoPendingMigrationsTest(TestCase):
+    """
+    Test if proper migrations are added and the models state is consistent.
+    It should make sure that no new migrations are created for this app,
+    when end-user runs `makemigrations` command.
+    """
+
+    def test_no_pending_migrations(self):
+        call_command("makemigrations", "silk", "--check", "--dry-run")
+
+    @override_settings(DEFAULT_AUTO_FIELD='django.db.models.BigAutoField')
+    def test_check_with_overridden_default_auto_field(self):
+        """
+        Test with `BigAutoField` set as `DEFAULT_AUTO_FIELD` - which is
+        default when generating proj with Django 3.2.
+        """
+        self.test_no_pending_migrations()
 
 
 class BaseProfileTest(TestCase):
--- a/silk/migrations/0009_explicit_id_fields.py
+++ b/silk/migrations/0009_explicit_id_fields.py
@@ -1,0 +1,20 @@
+from django.db import migrations, models
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('silk', '0008_sqlquery_analysis'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='profile',
+            name='id',
+            field=models.AutoField(primary_key=True, serialize=False),
+        ),
+        migrations.AlterField(
+            model_name='sqlquery',
+            name='id',
+            field=models.AutoField(primary_key=True, serialize=False),
+        ),
+    ]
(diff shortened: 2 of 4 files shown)


Archmonger:
I recommend commiting a new migration file alongside this PR.

silk/migrations/0009_explicit_id_fields.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations, models
+
+class Migration(migrations.Migration):
```
Archmonger, line 3: I would argue this is a good opportunity to remove makemigrations from the readme.

daadu:
@Archmonger Appropriate migration file added.

Also I have kept id filed of type `AutoField` only, since we should not be changing it to `BigAutoField` just because the default for new project will change, to it. If need to migrate to `BigAutoField` then it should have its own reasoning.

silk/migrations/0009_explicit_id_fields.py
daadu: :+1:  :done:

Archmonger:
This looks good to me but I'll defer the final review to Asif.

Archmonger (approved):

silk/models.py
```diff
@@ -231,6 +231,7 @@ def bulk_create(self, *args, **kwargs):
 
 
 class SQLQuery(models.Model):
+    id = models.AutoField(primary_key=True)
```
auvipy, line 234: instead of this 2  changes the app module with appconfig and default key would be appropriate.

auvipy (changes requested):
instead of ading explicit per model ID we can use appconfig

auvipy:
> This looks good to me but I'll defer the final review to Asif.

thanks, do initial review but please tag me for another round of review since you are new in this repo.

silk/migrations/0009_explicit_id_fields.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations, models
+
```
auvipy, line 2: when remvoe the explicit per moderl ID, and reintroduce appconfig, remove this file and regen a new migration and commit on this PR please.

project/tests/test_models.py
```diff
@@ -491,6 +492,25 @@ def test_delete_if_has_request(self):
         self.assertNotIn(self.obj, models.SQLQuery.objects.all())
 
 
+class NoPendingMigrationsTest(TestCase):
```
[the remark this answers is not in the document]
Archmonger: I'd argue this one is a moot point since PRs should almost always be squashed before entering the main branch.

silk/models.py
daadu: in that case this PR would be same as #495 ? Was it reverted only because it was "self-merged", I assumed it was because issue was with new "apps.py" file.

project/tests/test_models.py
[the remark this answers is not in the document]
auvipy: thats another point, squash merge or rebase merge both is ok.

silk/migrations/0009_explicit_id_fields.py
daadu: Migration would not be required in appconfig based approach.

silk/models.py
daadu: @Archmonger What additional components in this PR (except for README change, for which another PR could have been filed) would have been needed in it?

As per the review done, if we do those changes then this PR would be identical to #495.

I only wrote this PR because @auvipy told "yeah gonna revert this PR. we need different approch for handling this" he also reviewed with comment "bigAutoField will be used for new django projects autogen primary key, that should not effect this package, for the change we had we should generate new migrations which will need to be supplied with this package.". Therefore, I concluded that the approach we are taking is to have have explicit ids for two models (rest already use explicit ids).

After writing this PR, it is reviewed that need to go back to the #495, therefore I suggest you "revert the revert of #495". Or if we want to go with this (explicit ids in models) approach we should work on this PR. And about the README correct, I will file new PR, if that is agreed upon.

It would not make sense to write an identical PR. Hope that makes sense.

silk/migrations/0009_explicit_id_fields.py
auvipy: that would be ok

silk/models.py
auvipy: first of all you shouldn't have merged that PR. that needed thorough review. that is not how open source contributions are made. and no one told you to follow this approach. in django core your contributions would went through several round of review before actually get merged.

daadu:
I am opting out of contributing on this issue.

PR #503, opened 2021-10-11
Remove Make Migrations

Archmonger:
Based on Daadu's PR

Migration file not needed due to default_auto_field.


silk/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+class SilkAppConfig(AppConfig):
+    default_auto_field = 'django.db.models.AutoField'
+    name = 'silk'
```
auvipy, line 5: no new line

project/tests/test_app_config.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class TestAppConfig(TestCase):
+    """
+    Test if correct AppConfig class is loaded by Django.
+    """
+
+    def test_app_config_loaded(self):
+        silk_app_config = proj_apps.get_app_config("silk")
+        self.assertIsInstance(silk_app_config, SilkAppConfig)
```
auvipy, line 14: no new line.also can we use pytest style assert, if the project already use unitest stlyle assert then we can go as is

PR #521, opened 2021-11-02
Fix installation instructions in README

side2k:
Right now, instructions in README seem to be a bit inconsistent - code is [checking](https://github.com/jazzband/django-silk/blob/12a28e99ec0c629aeb379b0657ebfc56925f8d61/silk/profiling/profiler.py#L125) for `silk` being added to `INSTALLED_APPS` setting, and the README says about adding `silk.apps.SilkAppConfig` there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ MIDDLEWARE = [
 
 INSTALLED_APPS = (
     ...
-    'silk.apps.SilkAppConfig'
+    'silk'
 )
 ```
 


auvipy (changes requested):
https://github.com/jazzband/django-silk/blob/master/silk/apps.py#L4 this should work right?

side2k:
> https://github.com/jazzband/django-silk/blob/master/silk/apps.py#L4 this should work right?

@auvipy according to [docs](https://docs.djangoproject.com/en/3.2/ref/applications/#configuring-applications) this will be loaded automatically. It makes sense to use path to AppConfig class in `INSTALLED_APPS` only when there are more than one class in `apps` module.

auvipy (approved):

PR #532, opened 2021-12-26
Remove unneeded dependency Pygments

jayvdb:


tox.ini
```diff
@@ -11,6 +11,7 @@ DJANGO =
     2.2: dj22
     3.1: dj31
     3.2: dj32
+    4.0: dj40
```
auvipy, line 14: this is good addition, but why not in github actions as well

jayvdb: It is useful for locally testing, but I think dj40 is adequately covered with the django-main jobs in GHA.  Happy to add to GHA also if you think that it is better to get it done now.

auvipy: ok later

auvipy (approved):

Issue #371, opened 2019-10-28
Deadlock working with postgresql


mtali:
I don't know how to repricate this byt it seams silk cause deadlock on my database

```
django.db.utils.OperationalError: deadlock detected
DETAIL:  Process 2348 waits for ShareLock on transaction 28337148; blocked by process 2347.
Process 2347 waits for ShareLock on transaction 28337155; blocked by process 2348.
HINT:  See server log for query details.
CONTEXT:  while deleting tuple (367,10) in relation "silk_response"
```

nasirhjafri:
Closing as duplicate of #294

Issue #294, opened 2018-07-09
MySQL can deadlock on garbage collection


albertyw:
```
...
/silk/models.py", line 167, in save
    Request.garbage_collect(force=False)
/silk/models.py", line 152, in garbage_collect
    cls.objects.filter(start_time__lte=time_cutoff).delete()
...
/silk/sql.py", line 45, in execute_sql
    return self._execute_sql(*args, **kwargs)
...
django.db.utils.OperationalError: (1213, 'Deadlock found when trying to get lock; try restarting transaction')
```

meequz:
Experienced similar issue with PostgreSQL. Silk v.3.0.1, Django 1.11.13
Here are the details from Sentry:
```
TransactionRollbackError: deadlock detected
DETAIL:  Process 15422 waits for ShareLock on transaction 6038563; blocked by process 15420.
Process 15420 waits for AccessExclusiveLock on tuple (247,8) of relation 148876 of database 16384; blocked by process 15421.
Process 15421 waits for ShareLock on transaction 6038573; blocked by process 15422.
HINT:  See server log for query details.
CONTEXT:  while deleting tuple (404,8) in relation "silk_response"

  File "django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)

OperationalError: deadlock detected
DETAIL:  Process 15422 waits for ShareLock on transaction 6038563; blocked by process 15420.
Process 15420 waits for AccessExclusiveLock on tuple (247,8) of relation 148876 of database 16384; blocked by process 15421.
Process 15421 waits for ShareLock on transaction 6038573; blocked by process 15422.
HINT:  See server log for query details.
CONTEXT:  while deleting tuple (404,8) in relation "silk_response"

  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "django/utils/deprecation.py", line 138, in __call__
    response = self.process_request(request)
  File "silk/middleware.py", line 96, in process_request
    request_model = RequestModelFactory(request).construct_request_model()
  File "silk/model_factory.py", line 191, in construct_request_model
    body=body)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 394, in create
    obj.save(force_insert=True, using=self.db)
  File "silk/models.py", line 167, in save
    Request.garbage_collect(force=False)
  File "silk/models.py", line 152, in garbage_collect
[...]

Issue #265, opened 2018-01-26
DB Deadlock when stress testing with silk


selimt:
I got a deadlock error when testing with the latest version of silk (1.1.0).

Here is the DB output:

```
        Process 149: DELETE FROM "silk_response" WHERE "silk_response"."request_id" IN ('629f685a-26dc-41ca-ba9d-cc8af1998ef9', '8336effc-217c-47a7-8cc2-6944b8aede7f', 'aa12852b-5f5b-40fa-a647-cc3ddd2ef4ba', '2b2d5e84-e595-4ccf-b06c-78db181b4f24', '3f94c00d-f0c3-4021-8def-703e363e4f8f', 'e88b9162-0455-4e78-ba7b-029694b493d8')
        Process 150: DELETE FROM "silk_response" WHERE "silk_response"."request_id" IN ('629f685a-26dc-41ca-ba9d-cc8af1998ef9', '9ad245bb-4b6a-49cf-b32c-1a547ee8c11f', '93c50572-f6ec-416c-a9d6-42798f1631dd', '1ffd6a86-f93e-48f8-a593-86184e26c417', '3f94c00d-f0c3-4021-8def-703e363e4f8f')
        Process 151: DELETE FROM "silk_response" WHERE "silk_response"."request_id" IN ('629f685a-26dc-41ca-ba9d-cc8af1998ef9', '9ad245bb-4b6a-49cf-b32c-1a547ee8c11f', 'aa12852b-5f5b-40fa-a647-cc3ddd2ef4ba', '8336effc-217c-47a7-8cc2-6944b8aede7f', '2b2d5e84-e595-4ccf-b06c-78db181b4f24', '93c50572-f6ec-416c-a9d6-42798f1631dd', '1ffd6a86-f93e-48f8-a593-86184e26c417', '3f94c00d-f0c3-4021-8def-703e363e4f8f', 'e88b9162-0455-4e78-ba7b-029694b493d8')
[...]

PR #541, opened 2022-01-21
Added silk_request_garbage_collect command for out-of-band garbage collection.

nschlemm:
To avoid deadlock issues as mentioned e.g. in #265 #294 #371
Based on https://github.com/jazzband/django-silk/issues/265#issuecomment-705482767


hramezani:
Hi @nschlemm, Thanks for mentioning me in this patch and asking for a review.
Unfortunately, I am not involved in this project.
So, you may need to ask project leads for a review.

nschlemm:
@hramezani sorry, then github must definitely step up their game for suggesting reviewers ;)

silk/management/commands/silk_request_garbage_collect.py
```diff
@@ -0,0 +1,29 @@
...
+            "--max-requests",
+            default=SilkyConfig().SILKY_MAX_RECORDED_REQUESTS,
+            type=int,
+            help="Maximum number of requests to keep after garbage collection.",
+        )
+
+    def handle(self, *args, **options):
+        if "max_requests" in options:
+            max_requests = options["max_requests"]
+            SilkyConfig().SILKY_MAX_RECORDED_REQUESTS = max_requests
```
albertyw, line 22: Changing `SilkyConfig` in flight is kind of tricky because these changes aren't propagated back into the original django project's `settings.py`, possibly making it hard to debug issues.  Unfortunately, I think this is the only way to do so because `SilkyConfig` is a singleton and instantiated only once before this code will run.

albertyw (approved):

albertyw:
Even when triggered manually through this command, it's still possible for deadlocks to happen unless the database is receiving no other writes during the period.  This isn't going to guarantee fixing deadlocks.  There should be a full solution made to fix deadlocks, though the ability to manually trigger cleanup perhaps during low-traffic hours is still useful.

albertyw:
There are some rebase conflicts.  @nschlemm Please rebase and this can be merged.

PR #544, opened 2022-02-01
Update django version

### LINKED ISSUE #543: Update to django 4
Hey, I've been using the project for a while and I would like to contribute 😄 . I will try to open a new pull upgrading the django version to see if all tests are passing or not.

eduzen:
Hi, it's my first pull requests 😄 I'm trying to resolve #543

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
-        django-version: ['2.2', '3.1', '3.2', 'main']
+        django-version: ['2.2', '3.1', '3.2', '4.0', 'main']
 
     services:
       postgres:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
-        django-version: ['2.2', '3.1', '3.2', '4.0', 'main']
+        django-version: ['2.2', '3.2', '4.0', 'main']
 
     services:
       postgres:
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Silk is a live profiling and inspection tool for the Django framework. Silk inte
 
 Silk has been tested with:
 
-* Django: 2.2, 3.1, 3.2
+* Django: 2.2, 3.2
 * Python: 3.6, 3.7, 3.8, 3.9, 3.10
 
 ## Installation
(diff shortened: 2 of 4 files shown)


carltongibson (commented):
Hi @eduzen 👋 

So this isn't the right _fix_ here. 

See the `tox` file: https://github.com/jazzband/django-silk/blob/master/tox.ini

`tox` lets you run against various versions of different dependencies, and check they're all working. See https://tox.wiki/en/latest/

You'll see that the tests are already running against all supported versions of Django. 4.0 was added in b684b8ccf97a5553a1312ecbf588fc7523121a94.

In fact 3.1 is now EOL — end-of-life — so you can make a PR dropping testing against that. 


**With this change** I'm half expecting to see the Django 2.2 tests failing — as it can't install the exact requirement —  but they didn't seem to so I'd need to dig into the logs there to see exactly what gets installed. 🤔

eduzen:
thank you @carltongibson, I will take a look to the `tox.ini`

eduzen:
@carltongibson after some issues between pyenv and tox, I could finally manage to run the tests for all the versions and databases. 

Following your comments, I removed django 3.1 and python3.6 that is also now EOL.   Could you take a look to these changes :smile: ?

I think it would be useful to update the docs. Maybe I can create a issue for that and work on that. I think that the contribute part is outdated.  So we can prevent my mistake of not considering tox for running the tests.

.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
-        django-version: ['2.2', '3.1', '3.2', 'main']
+        django-version: ['2.2', '3.1', '3.2', '4.0', 'main']
```
carltongibson, line 13:
```
        django-version: ['2.2', '3.2', '4.0', 'main']
```

... would be all currently supported versions of Django.

carltongibson (commented):
I don't know who's project lead here currently... 🤔 ... but you'd need to get them to confirm dropping PY36 support (since Django 2.2 isn't quite EOL itself yet.) 

(Although it's only a couple of months away, so if you're doing the work...) 

Otherwise, looks good. 

I'd say, Yes! Do make a separate PR with the contributing docs changes you think — never better than when you've just stepped through it. (I'm happy to give those a look too, so @carltongibson me on that.)

carltongibson:
@eduzen See the checks: the PY36 entries haven't been correctly removed (you see the failures there...) 

Let me know if you need help lining that up. I've been using `tox-py` with great success, to simplify the GHA workflow. 

e.g. See the Channels workflow 

https://github.com/django/channels/blob/26fa1f366f82c83fd0b269155c68b7b71e852e26/.github/workflows/tests.yml#L25-L36

eduzen:
Thank you @carltongibson again for reviewing this. I put py36 back and remove django3.1.

carltongibson:
Thanks @albertyw ! 👍

eduzen:
thank you @carltongibson and @albertyw !

Issue #508, opened 2021-10-12
Upgrade jQuery to 3.6+


auvipy:

eduzen:
I updated jquery and jquery-ui, @auvipy do you have some opinion in the way we should test this change?

SebCorbin:
Although I would prefer we rely on no jQuery at all, which is the way to go, I'll have a look at the PR

PR #546, opened 2022-02-06
Update jquery to 3.6.0 and jquery-ui to 1.13.1 [#508]

eduzen:
#508 I'm trying to update jquery and jquery-ui . I did some tests with the test_app and I couldn't see any errors in the debugger console. I don't know if there are some tests for the part of the code in js.

If someone has a good idea of test for this is very welcome.


albertyw:
Yes, there's no JS tests for django-silk.  I checked through the JS code in https://github.com/albertyw/django-silk/tree/master/silk/static/silk/js/ and I don't see anything that would cause issues with upgrading jquery and jquery-ui (I'm actually not sure if the latter is even used or if it can be safely deleted) so this PR looks good.

albertyw:
Yes, there's no JS tests for django-silk.  I checked through the JS code in https://github.com/albertyw/django-silk/tree/master/silk/static/silk/js/ and I don't see anything that would cause issues with upgrading jquery and jquery-ui (I'm actually not sure if the latter is even used or if it can be safely deleted) so this PR looks good.

albertyw (approved):

albertyw (approved):

PR #550, opened 2022-02-15
Some tests misusing assertTrue for comparisons fix

### LINKED ISSUE #549: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments, should use `assertEqual` for that.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is assertEqual. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/jazzband/django-silk/blob/8f8d159a4209f4c294aba16bce3b1ca5789dc2d2/project/tests/test_silky_profiler.py#L59
https://github.com/jazzband/django-silk/blob/8f8d159a4209f4c294aba16bce3b1ca5789dc2d2/project/tests/test_silky_profiler.py#L85

I found this issue automatically, see other issues [here](https://codereview.doctor/jazzband/django-silk)

code-review-doctor:
Fixes #549

### DIFF
--- a/project/tests/test_silky_profiler.py
+++ b/project/tests/test_silky_profiler.py
@@ -56,7 +56,7 @@ class TestProfilertContextManager(TestCase):
             sleep(0.1)
 
     def test_one_object(self):
-        self.assertTrue(len(DataCollector().profiles), 1)
+        self.assertEqual(len(DataCollector().profiles), 1)
 
     def test_name(self):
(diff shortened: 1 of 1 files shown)


eduzen:
hey @code-review-doctor good catch, LGTM.

PR #556, opened 2022-03-02
Pin dependencies in requirements.txt

albertyw:
As suggested in #548, this pins dependencies in requirements.txt.  These packages are used for development and testing of django-silk and will not affect upstream Django projects that use django-silk.  

This set of dependencies is not compatible with Python 3.6, hence this PR depends on #554.

All versions are the latest for each dependency, with the exception of `networkx` being pinned to 2.6.  Networkx's latest version is [2.7](https://networkx.org/documentation/stable/release/release_2.7.html) but 2.7 drops support for Python 3.7 which is still officially supported by the Python Software Foundation and by django-silk.

This also depends on #555.


setup.py
```diff
@@ -22,12 +22,11 @@
         'Environment :: Web Environment',
         'Framework :: Django',
         'Framework :: Django :: 2.2',
-        'Framework :: Django :: 3.1',
```
Archmonger, line 25: Is there a reason for us to drop 3.1 support, but keep 2.2?

albertyw: Django 2.2 is an LTS release and supported until April.  Django 3.1 EOLed back in December.

Source: https://www.djangoproject.com/download/#supported-versions

Archmonger (approved):

hramezani (approved):

Issue #530, opened 2021-12-15
How to configure EXPLAIN format with silk?


iamkarlson:
Currently, silk uses just EXPLAIN without any flags for profiling posrgres sql queries. Certain tools like https://app.pgmustard.com require JSON to be supplied so they can work the best. How would it be possible to use additional flags like ANALYZE, VERBOSE, BUFFERS, FORMAT JSON with current implementation of the silk profiler?

Archmonger:
Implementing a custom `SILKY_PYTHON_PROFILER_FUNC = ...` would be the way to go for your scenario. 

You can also PR something like a `SILKY_EXPLAIN_FLAGS` variable as well.

baruchoxman:
@iamkarlson @Archmonger I've implemented this functionality, any chance you could review and possibly approve my PR?

baruchoxman:
@iamkarlson @albertyw Should we mark this one as "resolved" following my fix being merged?

iamkarlson:
Sorry for leaving you out. We decided to stop using silk as it's heavily hurting our database usage. if we get back - I will certainly try it out! Thanks!

PR #558, opened 2022-03-07
Fixing #530 - Adding support for SILKY_EXPLAIN_FLAGS

baruchoxman:
Adding support for additional flags that can be passed as part of the "analyze" query, for more explicit analysis.

### DIFF
--- a/silk/sql.py
+++ b/silk/sql.py
@@ -38,10 +38,11 @@ def _explain_query(connection, q, params):
                 )
             except ValueError as error:
                 error_str = str(error)
-                if error_str.startswith("Unknown options:") and "analyze" in error_str:
+                if error_str.startswith("Unknown options:"):
                     Logger.warning(
-                        "Database does not support analyzing queries. "
-                        "SILKY_ANALYZE_QUERIES option will be ignored"
+                        "Database does not support analyzing queries with provided params. %s."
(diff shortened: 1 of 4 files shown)


silk/sql.py
```diff
@@ -34,10 +34,11 @@ def _explain_query(connection, q, params):
             # Work around some DB engines not supporting analyze option
             try:
                 prefix = connection.ops.explain_query_prefix(
-                    analyze=True
+                    analyze=True, **(SilkyConfig().SILKY_EXPLAIN_FLAGS or {})
                 )
             except ValueError as error:
-                if str(error) == "Unknown options: analyze":
+                error_str = str(error)
+                if error_str.startswith("Unknown options:") and "analyze" in error_str:
```
liorzmetis: When error occurred it would be better to show which option was a problem and append the error to the logger.

liorzmetis (changes requested):
looks good

silk/sql.py
baruchoxman: Thanks, fixed

liorzmetis (approved):

Archmonger (approved):

albertyw:
Merged.  Thanks for the PR.

PR #568, opened 2022-05-28
Lint the codes with black and isort.

gholamiayub:


tony:
@gholamiayub Any reason why this was closed? black + isort makes sense to me

PR #590, opened 2022-08-12
Use contextlib in test_profile_parser

s-t-e-v-e-n-k:
Since the project no longer declares a requirement of contextlib2, and
the lowest version of Python supported is 3.7, we should remove the use
of contextlib2. Switch the import to contextlib.


albertyw:
Can you also edit `requirements.txt` and remove the `contextlib2` dependency there too?

PR #596, opened 2022-09-12
Add support for storages, that don't implement full filesystem path

mizi:
This PR fixes https://github.com/jazzband/django-silk/issues/208

### DIFF
--- a/silk/collector.py
+++ b/silk/collector.py
@@ -1,5 +1,6 @@
 import cProfile
 import logging
+import marshal
 import pstats
 from io import StringIO
 from threading import local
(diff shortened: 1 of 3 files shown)


silk/collector.py
```diff
@@ -144,8 +145,8 @@ def finalise(self):
 
             if SilkyConfig().SILKY_PYTHON_PROFILER_BINARY:
                 file_name = self.request.prof_file.storage.get_available_name(f"{str(self.request.id)}.prof")
-                with open(self.request.prof_file.storage.path(file_name), 'w+b') as f:
-                    ps.dump_stats(f.name)
+                with self.request.prof_file.storage.open(file_name, 'w+b') as f:
+                    marshal.dump(ps.stats, f)
```
albertyw, line 149: Are there any security ramifications of using `marshal.dump` on `pstats`?  My guess is no but I'm not that familiar with what access an external user might get to `pstats.Stats`.

For backwards compatibility, would it not be better to continue to use `dump_stats` and have it write to a file object (even a fake file object)?

mizi: @albertyw Here's the original `pstats.Stats.dump_stats`
```python
    def dump_stats(self, filename):
        """Write the profile data to a file we know how to load back."""
        with open(filename, 'wb') as f:
            marshal.dump(self.stats, f)
```
I only moved this `marshal.dump()` call outside, to avoid creating a temporary file just for `ps.dump_stats()` to open it again using a filename. Creating a temporary file has some performance penalty.
Another solution would be to subclass `pstats.Stats` and add a new method that accepts a file object instead of a filename. But it'll be essentially the same code.

albertyw (approved):

PR #609, opened 2022-10-26
Highlight non third-party code in traceback

SebCorbin:


silk/views/sql_detail.py
```diff
@@ -55,7 +58,8 @@ def get(self, request, *_, **kwargs):
             'pos': pos,
             'line_num': line_num,
             'file_path': file_path,
-            'analysis': analysis
+            'analysis': analysis,
+            'virtualenv_path': os.environ.get('VIRTUAL_ENV'),
```
albertyw, line 62: VIRTUAL_ENV is not guaranteed to be set, so this may also return `None`.  The template file would then try to run `None not in "string"` which could result in a TypeError.

Also, is it possible for VIRTUAL_ENV to be an absolute path but the traceback to be relative and not include substrings with VIRTUAL_ENV?

Are you sure this is going to work?  There should be a test case for if `VIRTUAL_ENV` is not set.

SebCorbin: - Added tests and default value of `''` (even if the `not in` operator handles comparison with `None`)
- VIRTUAL_ENV is not guaranted to be set [as per docs](https://docs.python.org/3/library/venv.html#how-venvs-work), but if it is, it's an absolute path
  > When a virtual environment has been activated, the VIRTUAL_ENV environment variable is set to the path of the environment. Since explicitly activating a virtual environment is not required to use it, VIRTUAL_ENV cannot be relied upon to determine whether a virtual environment is being used.
- This is a non-intrusive enhancement: no virtualenv means no CSS class added, and if the path are mixed between relative and absolute (which I doubt will happen), either all lines will be hightlighted or none will

albertyw (approved):

PR #631, opened 2022-11-20
Use Django 2.0 execute_wrapper()

### LINKED ISSUE #214: Consider using Django 2.0 `execute_wrapper()`
Silk currently uses monkeypatching to install its wrapper around `SQLCompiler.execute_sql`. Django 2.0 provides a new API for installing wrappers around SQL executions, and it may be preferable to use this API when available.

Points arising:
* The new API works at a lower level, that of the connection rather than `SQLCompiler`. I'm not sure what the implications are.
* The new API installs the wrapper only in the current thread and in a context-manager which removes it on exit, these seem to be improvements over the current monkey-patching.
* There is an intention to make the new API available for older versions of Django, via a 3rd-party package. So, when checking for its availability, the right check is not  a Django version check, but rather `hasattr(connection, 'execute_wrappers')`.

SebCorbin:


ulgens:
@SebCorbin Can you please give some details about this PR and its purpose? Thank you.

SebCorbin:
@ulgens yes, this related issue and last comment https://github.com/jazzband/django-silk/issues/214#issuecomment-1321204137

SebCorbin:
After few days of real-world testing, the `ready()` method is not development-proof with the `runserver` command. I'll update this PR to switch to classic context manager on the middleware.

SebCorbin:
Another option: rely on `connection_created` signal

PR #637, opened 2023-01-01
632 Delete Requests

### LINKED ISSUE #632: Allow to delete some requests
In development, I let silk live in the background and I often spawn its summary page to see page that takes the most time/queries. 

Now I sometimes use `pdb` on some requests so the time calculated for these is not relevant, and it would be nice to have a button to remove them from the DB so that the summary page can be updated.

ashfaq1934:
- Adds a button to each request card and detail page which allows you to delete individual request without having to click clear db.
- Resolves #632


silk/views/request_detail.py
```diff
@@ -40,3 +40,13 @@ def get(self, request, request_id):
...
+
+    @method_decorator(login_possibly_required)
+    @method_decorator(permissions_possibly_required)
+    def post(self, request, request_id):
+        silk_request = Request.objects.get(pk=request_id)
+        try:
+            silk_request.delete()
+        except Request.DoesNotExist:
+            return HttpResponseRedirect(request.headers.get('Referer', '/silk'))
+        return redirect('/silk')
```
gopar, line 52: Probably better to call `reverse("summary")` here and the line above as well to not hardcode `/silk` path 
(Not everyone will use the `/silk` in their paths)

ashfaq1934: Thanks for the feedback. I've now removed the hard-coded path and replaced it with `reverse('silk:summary')`.

project/tests/test_delete_request.py
```diff
@@ -0,0 +1,20 @@
...
+        self.client = Client()
+        self.request = Request()
+        self.request.path = reverse('silk:requests')
+        self.request.method = 'get'
+        self.request.body = b'a' * 1000
+        self.request.save()
+
+    def test_delete(self):
+        response = self.client.post(reverse('silk:request_detail', kwargs={'request_id': str(self.request.id)}))
+        self.assertEqual(response.status_code, HTTPStatus.FOUND)
```
gopar, line 20: Thanks for the tests! A few nitpicks:
- Can you make sure that this re-routes to the summary page and the object is deleted. (Bonus for test that checks the redirection for header Referer)
- Another test that checks when we pass in a non-existing request

PR #641, opened 2023-01-12
Switch from request META to request headers

albertyw:
As recommended by https://github.com/adamchainz/django-upgrade#httprequestheaders, switch from `request.META` to `request.headers`.  This slightly simplifies several test cases by not having to convert between META and headers (see https://docs.djangoproject.com/en/dev/ref/request-response/#httprequest-objects for more information).  However, headers are case insensitive which introduces some case-changes.  

All of the changes in this PR are limited to tests, except for a very minor change when generating curls.

This is a followup to https://github.com/jazzband/django-silk/pull/640#issuecomment-1375269162

### DIFF
--- a/silk/model_factory.py
+++ b/silk/model_factory.py
@@ -45,7 +45,7 @@ class DefaultEncoder(json.JSONEncoder):
 
 
 def _parse_content_type(content_type):
-    """best efforts on pulling out the content type and encoding from Content-Type header"""
+    """best efforts on pulling out the content type and encoding from content-type header"""
     char_set = None
     if content_type.strip():
         splt = content_type.split(';')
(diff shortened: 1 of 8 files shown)


silk/model_factory.py
```diff
@@ -45,7 +45,7 @@ def default(self, o):
 
 
 def _parse_content_type(content_type):
-    """best efforts on pulling out the content type and encoding from Content-Type header"""
+    """best efforts on pulling out the content type and encoding from content-type header"""
     try:
         content_type = content_type.strip()
     except AttributeError:
```
SebCorbin, line 51: It seems this line is no longer covered, but I can't think of a situation where `content_type` would not be a string and thus would trigger an `AttributeError`, maybe we can remove this try-catch?

SebCorbin (approved):

PR #652, opened 2023-04-06
Allow option to delete profiles

viralj:
Added a new config option that will allow user to clear profiler profiles from storage when Clear DB is called. This is helpful functionality especially in development environment where frequent DB cleaning is done.


viralj (commented):
Looks like correct update.

viralj (commented):
Added new test case

viralj:
All checks have passed and the updated code implements helpful functionality.

albertyw (approved):

Issue #650, opened 2023-03-13
Calling explain() causes a syntax error


mapavia:
```
ProgrammingError at /admin/user/user/
syntax error at or near "EXPLAIN"
LINE 1: EXPLAIN EXPLAIN SELECT "user"."password", "user"."is_superus...
```

Can I explicitly add a check on `silk/sql.py`

```

        # currently we cannot use explain() method
        # for queries other than `select`
        if not q.startswith("EXPLAIN"):
            prefixed_query = f"{prefix} {q}"
            with connection.cursor() as cur:
                cur.execute(prefixed_query, params)
                result = _unpack_explanation(cur.fetchall())
                return '\n'.join(result)
    return None
```

stereodamage:
Possible duplicate of #597

danmaxis:
I am currently having this issue with django-silk==5.3.2.
It is conflicting with some explain queries from djangoql==0.18.1.

These EXPLAIN queries are arriving in the q variable and it throws an exception on running the sql with 'EXPLAIN EXPLAIN...'

![Image](https://github.com/user-attachments/assets/e634da46-0110-4123-b84e-0ae6e95c0c48)

PR #654, opened 2023-04-23
Fix double EXPLAIN when calling explain on queryset

### LINKED ISSUE #597: Running queryset.explain will result in an error
Using djangoql as example which does run queryset.explain will result in a ProgrammingError.
Reproducing would be easy by adding a queryset.explain in a view or installing [djangoql](https://github.com/ivelum/djangoql) and try to preform a search in the Django admin.

The generated SQL will contains two explains 

```
syntax error at or near "EXPLAIN"
LINE 1: EXPLAIN EXPLAIN SELECT ...
```

This is caused by `_explain_query` and exactly this part
```python
        # currently we cannot use explain() method
        # for queries other than `select`
        prefixed_query = f"{prefix} {q}"
        with connection.cursor() as cur:
            cur.execute(prefixed_query, params)
            result = _unpack_explanation(cur.fetchall())
            return '\n'.join(result)
```

A possible way to fix this is to check if the SQL starts with explain and skip the check/analyze.

```python
        # currently we cannot use explain() method
        # for queries other than `select`
        # Skip explain queries
        if q.lower().startswith('explain'):
            return None
        prefixed_query = f"{prefix} {q}"
        with connection.cursor() as cur:
            cur.execute(prefixed_query, params)
            result = _unpack_explanation(cur.fetchall())
            return '\n'.join(result)
```

If that's fine I can prepare a PR for this.

Thank you.

stereodamage:
Closes #597, #650.

There's also an idea to wrap call to database into `try..except` and throw something like `ProgrammingError` (according to [PEP 249](https://peps.python.org/pep-0249/#programmingerror)) or custom error based on ProgrammingError.


pawelzar:
Hey, I just encountered this issue too. I think the changes look good but test coverage is failing. I can help you add some unit tests if you'd like @stereodamage.

silk/sql.py
```diff
@@ -53,7 +54,11 @@ def _explain_query(connection, q, params):
 
         # currently we cannot use explain() method
         # for queries other than `select`
-        prefixed_query = f"{prefix} {q}"
+        if re.match(f"^{prefix} .*", q):
```
pawelzar, line 57: What do you think about using just `str.startswith()` instead?
```suggestion
        if q.startswith(prefix):
```

stereodamage: Yeah, that will work!

stereodamage:
@pawelzar that would be nice of you, I appreciate that. Didn't really had much time to look into it again and write new test myself. Anything needed from my side?

stereodamage:
@pawelzar finally found some time to fix tests and review changes, looks good now for me, can you take a look as well?

stereodamage:
@albertyw maybe you could take a look if you have time?

silk/sql.py
```diff
@@ -53,7 +53,11 @@ def _explain_query(connection, q, params):
 
         # currently we cannot use explain() method
         # for queries other than `select`
-        prefixed_query = f"{prefix} {q}"
+        if q.startswith(prefix):
```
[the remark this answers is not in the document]
stereodamage: Honestly, I'm not 100% sure. `prefix` supposed to be uppercase (django/db/backends/base/operations.py:760), and `q` - _could_ be lowercase, so yeah, let's stick with `.upper()` them both just to be sure.

albertyw:
@jcardali There are some more urgent fixes which need to go out (i.e. #819).

@stereodamage I took a stab at fixing the merge conflict on this.

albertyw (approved):

PR #669, opened 2023-08-06
Fix deprecation warning for get_storage_class

### LINKED ISSUE #656: RemovedInDjango51Warning for get_storage_class with Django 4.2
Using django-silk with Django 4.2 gives the following warning:

```
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/local/lib/python3.11/site-packages/silk/models.py", line 28, in <module>
    silk_storage = get_storage_class(SilkyConfig().SILKY_STORAGE_CLASS)()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/files/storage/__init__.py", line 32, in get_storage_class
    warnings.warn(GET_STORAGE_CLASS_DEPRECATED_MSG, RemovedInDjango51Warning)
django.utils.deprecation.RemovedInDjango51Warning: django.core.files.storage.get_storage_class is deprecated in favor of using django.core.files.storage.storages.
```

albertyw:
This PR should fix the deprecation warning for `get_storage_class` by trying to import and use the `storages` class instead.  However, the latter is not an exact replacement for the former, requiring a different configuration option to be provided under the `STORAGE` instead of the `SILKY_STORAGE_CLASS` setting.  This makes this PR a forwards incompatible change requiring a minor semver bump.

Fixes #656

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,16 @@ When enabled, a graph visualisation generated using [gprof2dot](https://github.c
 A custom storage class can be used for the saved generated binary `.prof` files:
 
 ```python
+# For Django >= 4.2 and Django-Silk >= 5.1.0:
+# See https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-STORAGES
+STORAGES = {
+    'SILKY_STORAGE': {
+        'BACKEND': 'path.to.StorageClass',
+    },
+    # ...
+}
+
+# For Django < 4.2 or Django-Silk < 5.1.0
 SILKY_STORAGE_CLASS = 'path.to.StorageClass'
 ```
 
--- a/silk/models.py
+++ b/silk/models.py
@@ -5,7 +5,7 @@ import re
 from uuid import uuid4
 
 import sqlparse
-from django.core.files.storage import get_storage_class
+from django.conf import settings
 from django.db import models, transaction
 from django.db.models import (
     BooleanField,
@@ -25,7 +25,20 @@ from django.utils.safestring import mark_safe
 from silk.config import SilkyConfig
 from silk.utils.profile_parser import parse_profile
 
-silk_storage = get_storage_class(SilkyConfig().SILKY_STORAGE_CLASS)()
+try:
+    # New in Django 4.2
+    from django.core.files.storage import storages
+    from django.core.files.storage.handler import InvalidStorageError
+    try:
+        silk_storage = storages['SILKY_STORAGE']
+    except InvalidStorageError:
+        from django.utils.module_loading import import_string
+        storage_class = SilkyConfig().SILKY_STORAGE_CLASS or settings.DEFAULT_FILE_STORAGE
+        silk_storage = import_string(storage_class)()
+except ImportError:
+    # Deprecated since Django 4.2, Removed in Django 5.1
+    from django.core.files.storage import get_storage_class
+    silk_storage = get_storage_class(SilkyConfig().SILKY_STORAGE_CLASS)()
 
 
 # Seperated out so can use in tests w/o models


README.md
```diff
@@ -170,6 +170,15 @@ When enabled, a graph visualisation generated using [gprof2dot](https://github.c
 A custom storage class can be used for the saved generated binary `.prof` files:
 
 ```python
+# For Django >= 4.2 and Django-Silk >= 5.1.0:
+STORAGES = {
+    'SILKY_STORAGE': {
+        'BACKEND': 'path.to.StorageClass',
+    },
+    # ...
+}
```
50-Course, line 179: 👍🏼

50-Course (approved):
Thanks for this PR! It's great to see that the deprecation warning for `get_storage_class` being addressed in earnest.

I noticed that the PR introduces a forwards incompatibility by requiring a different configuration option (STORAGE) to be used instead of `SILKY_STORAGE_CLASS` - a major change that will require users to update their configuration files. 

I'm also curious about how the `storages` class compares to the old `get_storage_class` approach in terms of functionality and performance.

Overall, I think this is a good PR that addresses an important issue.

albertyw:
The `STORAGES` config is defined by the Django framework itself and its predecessor `get_storage_class` is removed starting Django 5.1 so there's not much of an alternative to replacing `get_storage_class`.  All users, whether using django-silk or not will be eventually forced to abandon `get_storage_class` and therefore `SILKY_STORAGE_CLASS` and set up a `STORAGES` config.  Nevertheless, this change is a no-op for users using the default `silk.storage.ProfilerResultStorage` due to the try/except falling back to that storage if `storages['SILKY_STORAGE']` is not specified.

In terms of performance, comparing the new [storages object](https://github.com/django/django/blob/5.0/django/core/files/storage/handler.py#L33-L45) against the [get_storage_class](https://github.com/django/django/blob/5.0/django/core/files/storage/__init__.py#L31) implementation, it seems that the former would be more performant due to it caching storages.  Since django-silk is initializing `silk_storage` once on initialization rather than in the request-serving path, I don't think performance is a significant consideration though.  Either storages configuration gives equivalent functionality; the change only affects configuration format.

50-Course:
> The `STORAGES` config is defined by the Django framework itself and its predecessor `get_storage_class` is removed starting Django 5.1 so there's not much of an alternative to replacing `get_storage_class`. All users, whether using django-silk or not will be eventually forced to abandon `get_storage_class` and therefore `SILKY_STORAGE_CLASS` and set up a `STORAGES` config. Nevertheless, this change is a no-op for users using the default `silk.storage.ProfilerResultStorage` due to the try/except falling back to that storage if `storages['SILKY_STORAGE']` is not specified.
> 
> In terms of performance, comparing the new [storages object](https://github.com/django/django/blob/5.0/django/core/files/storage/handler.py#L33-L45) against the [get_storage_class](https://github.com/django/django/blob/5.0/django/core/files/storage/__init__.py#L31) implementation, it seems that the former would be more performant due to it caching storages. Since django-silk is initializing `silk_storage` once on initialization rather than in the request-serving path, I don't think performance is a significant consideration though. Either storages configuration gives equivalent functionality; the change only affects configuration format.

Thanks for the explanation. Coming back to this in a few.

PR #674, opened 2023-09-08
Handle case-insensitive sensitive headers

### LINKED ISSUE #673: SILKY_HIDE_COOKIES does not hide cookie header in different case
The SILKY_HIDE_COOKIES setting does not hide the cookie header if it is written in a different case.

**Current behavior:**
When using the `SILKY_HIDE_COOKIES` setting, the cookie header is not hidden if it is written in a different case compared to the specified name in the code in upper case 
**Expected behavior:**
The `SILKY_HIDE_COOKIES` setting should hide the cookie header, regardless of the case in which it is written. Currently, it only works if the case matches exactly.

**Steps to reproduce:**
1. Set the `SILKY_HIDE_COOKIES` setting to True.
2. Send a request with a 'cookie' header in lower case.
3. Inspect the request headers on the server side, and observe that the 'COOKIE' header is present.

**Refs**
- `COOKIE` in upper case is hardcoded [here](https://github.com/jazzband/django-silk/blob/master/silk/model_factory.py#L79)

shtimn:
fixes #673

### DIFF
--- a/silk/model_factory.py
+++ b/silk/model_factory.py
@@ -72,7 +72,7 @@ class RequestModelFactory:
         if SilkyConfig().SILKY_HIDE_COOKIES:
             sensitive_headers.add('cookie')
 
-        sensitive_headers = map(str.lower, sensitive_headers)
+        sensitive_headers = set(map(str.lower, sensitive_headers))
 
         headers = {}
         for k, v in self.request.headers.items():
--- a/silk/model_factory.py
+++ b/silk/model_factory.py
@@ -66,19 +66,21 @@ class RequestModelFactory:
         """
         From Django docs (https://docs.djangoproject.com/en/2.0/ref/request-response/#httprequest-objects):
         """
-        headers = {}
-        sensitive_headers = {'authorization'}
+        sensitive_headers = set(SilkyConfig().SILKY_SENSITIVE_KEYS)
 
+        sensitive_headers.add('authorization')
+        if SilkyConfig().SILKY_HIDE_COOKIES:
+            sensitive_headers.add('cookie')
+
+        sensitive_headers = map(str.lower, sensitive_headers)
+
+        headers = {}
         for k, v in self.request.headers.items():
+            k = k.lower()
             if k in sensitive_headers:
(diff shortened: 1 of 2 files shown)


silk/model_factory.py
```diff
@@ -66,19 +66,20 @@ def encoded_headers(self):
         """
         From Django docs (https://docs.djangoproject.com/en/2.0/ref/request-response/#httprequest-objects):
         """
-        headers = {}
-        sensitive_headers = {'authorization'}
+        sensitive_headers = SilkyConfig().SILKY_SENSITIVE_KEYS
```
kekekekule, line 69: maybe it's worth using 
```
set(SilkyConfig().SILKY_SENSITIVE_KEYS)
```
just for right type guarantee

silk/model_factory.py
```diff
@@ -66,19 +66,20 @@ def encoded_headers(self):
...
+
+        sensitive_headers.add('authorization')
+        if SilkyConfig().SILKY_HIDE_COOKIES:
+            sensitive_headers.add('cookie')
+
+        sensitive_headers = map(str.lower, sensitive_headers)
 
+        headers = {}
         for k, v in self.request.headers.items():
             if k in sensitive_headers:
```
kekekekule, line 79: let's do k.lower() maybe to avoid issues, if smth changes in Django?

kekekekule (changes requested):

silk/model_factory.py
shtimn: Oh, really
I missed that

silk/model_factory.py
shtimn: you're right
done

kekekekule (approved):

shtimn:
Hi @albertyw,

I have made some improvements to the handling of sensitive headers in the `RequestModelFactory.encoded_headers` method. 
Could you please take a moment to review the changes ? 

Thanks!

albertyw (approved):

shtimn:
@albertyw,  I appreciate the merge 😊

Could you please let me know when the new version is scheduled?

albertyw:
New version will be released with #678

Issue #682, opened 2023-10-24
[Python 3.12] ValueError: Another profiling tool is already active


Elendiar:
In python 3.12 sometimes this exception raises, usually immediately after the server starts in the first request.
`SILKY_PYTHON_PROFILER = False` seems to fix it.
```
File "/backend/python/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
  response = get_response(request)

File "/backend/python/lib/python3.12/site-packages/silk/middleware.py", line 70, in __call__
  self.process_request(request)

File "/backend/python/lib/python3.12/site-packages/silk/middleware.py", line 121, in process_request
  DataCollector().configure(request_model, should_profile=should_profile)

File "/backend/python/lib/python3.12/site-packages/silk/collector.py", line 95, in configure
  self.local.pythonprofiler.enable()
ValueError: Another profiling tool is already active
```

https://github.com/python/cpython/issues/110770#issuecomment-1759986100

PR #686, opened 2023-11-25
Support and Test Against Django 5

albertyw:
Depends on #669 
Depends on #685 
Depends on #687 
Depends on #692 
Depends on #699

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
50-Course, line 1: Looks good to me 👍🏼

50-Course (approved):
LGTM. Approving MR, since it does not introduce major breaking change.

albertyw:
Looking at logs, it seems the tests are failing due to `ValueError: Another profiling tool is already active` which was reported in #682

50-Course (approved):
LGTM, initially worried about the move from release candidate to stable, just confirmed it from the docs.

Pardon me, still work with older versions over here :)

PR #692, opened 2023-12-30
Gracefully error out when there are concurrent profilers

### LINKED ISSUE #682: [Python 3.12] ValueError: Another profiling tool is already active

albertyw:
Fixes #682 
Unblocks #686

### DIFF
--- a/project/tests/test_collector.py
+++ b/project/tests/test_collector.py
@@ -1,3 +1,6 @@
+import cProfile
+import sys
+
 from django.test import TestCase
 from tests.util import DictStorage
 
@@ -46,6 +49,18 @@ class TestCollector(TestCase):
                 content = f.read()
                 self.assertTrue(content)
                 self.assertGreater(len(content), 0)
+
+    def test_configure_exception(self):
+        other_profiler = cProfile.Profile()
+        other_profiler.enable()
+        collector = DataCollector()
+        collector.configure()
+        other_profiler.disable()
+        if sys.version_info >= (3, 12):
+            self.assertEqual(collector.local.pythonprofiler, None)
+        else:
+            self.assertIsNotNone(collector.local.pythonprofiler)
+            collector.stop_python_profiler()
 
     def test_profile_file_name_with_disabled_extended_file_name(self):
         SilkyConfig().SILKY_PYTHON_PROFILER_EXTENDED_FILE_NAME = False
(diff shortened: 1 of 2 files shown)


50-Course (commented):
Hello again @albertyw,

this patch permanently fixes the `Django==5.1` & `python^=3.12` pipeline issue?

50-Course (approved):
Update: Great progress! The "Gracefully error out when there are concurrent profilers" job (#722) is now passing, which has also unblocked #686.

robinchow:
I'd like to point out while this stop silk from interfering with processing a request, it also ends up *not profiling* that particular request. Prior to Python 3.11, the effect was similar but different (it would disable/stop the first profiler, and then start this current one). Neither outcome seems great, especially since this problem will happen when Django is processing multiple long-running requests at the same time (I've made a minimal repo [here](https://github.com/robinchow/django-silk-multithread.git) to highlight this issue)

I think at a minimum, the documentation should tell users that this issue exists. In terms of a proper fix, I'm not sure at the moment if that's possible without using another profiler that properly supports multithreaded applications (such as https://github.com/sumerc/yappi)

50-Course:
Thank you for pointing out this oversight, @robinchow. I'd take a look into [yappi](https://github.com/sumerc/yappi)

albertyw:
> Prior to Python 3.11, the effect was similar but different (it would disable/stop the first profiler, and then start this current one)

3.11 or 3.12?  This code gives an error in python 3.12 but no error in 3.11:

```python
import cProfile
p1 = cProfile.Profile()
p1.enable()
p2 = cProfile.Profile()
p2.enable()
```

robinchow:
@albertyw ah yes, apologies, prior to Python 3.12.

PR #694, opened 2023-12-31
Release version 5.1.0

albertyw:
Depends on #692 
Depends on #686 
Depends on #693


50-Course (commented):
When is this scheduled for release @albertyw?

albertyw:
There is no calendar schedule for releases of django-silk, but I would like to have this release support Django 5 (#686) and python 3.12 (#699) rather than those requiring a second release.

PR #695, opened 2024-01-07
Clarify max body size units and format readme

### LINKED ISSUE #300: Max body size units
I have a question regarding the `SILKY_MAX_REQUEST_BODY_SIZE` and `SILKY_MAX_RESPONSE_BODY_SIZE` settings.

The README example is following:

```
SILKY_MAX_REQUEST_BODY_SIZE = -1  # Silk takes anything <0 as no limit
SILKY_MAX_RESPONSE_BODY_SIZE = 1024  # If response body>1024kb, ignore
```

From that comment I inferred that the unit of the setting should be kB, However then I took a look at the source code and `sys.getsizeof` is used to get the size of the bodies. Doesn't that function return number of bytes and not number of kilobytes? 

In what unit should the setting be set? Kilobytes or bytes? In anycase I'd be happy to send a PR making this clearer in the README or updating the comparison in the ModelFactories to take the setting as kB...

AbhiMisRaw:
It's my first contribution, as @SebCorbin said, It's an docs issue, so I've updated it `BODY -> BYTE`

Fixes #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,25 +13,25 @@ Silk is a live profiling and inspection tool for the Django framework. Silk inte
 
 ## Contents
 
-* [Requirements](#requirements)
-* [Installation](#installation)
-* [Features](#features)
-* [Configuration](#configuration)
-  * [Authentication/Authorisation](#authenticationauthorisation)
-  * [Request/Response bodies](#requestresponse-bodies)
-  * [Meta-Profiling](#meta-profiling)
-  * [Recording a fraction of requests](#recording-a-fraction-of-requests)
-  * [Limiting request/response data](#limiting-requestresponse-data)
-  * [Clearing logged data](#clearing-logged-data)
-* [Contributing](#contributing)
-  * [Development Environment](#development-environment)
+- [Requirements](#requirements)
+- [Installation](#installation)
+- [Features](#features)
+- [Configuration](#configuration)
+  - [Authentication/Authorisation](#authenticationauthorisation)
+  - [Request/Response bodies](#requestresponse-bodies)
+  - [Meta-Profiling](#meta-profiling)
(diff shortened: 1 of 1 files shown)


50-Course (approved):
Hey @AbhiMisRaw, welcome to the `django-silk`!  Awesome first contribution—thanks for catching that docs issue and making the update to `BYTE`. It looks perfect!

albertyw (approved):

README.md
```diff
@@ -429,8 +426,8 @@ no matter how large. If Silk is used in production under heavy volume with large
 a huge impact on space/time performance. This behaviour can be configured with the following options:
 
 ```python
-SILKY_MAX_REQUEST_BODY_SIZE = -1  # Silk takes anything <0 as no limit
-SILKY_MAX_RESPONSE_BODY_SIZE = 1024  # If response body>1024 bytes, ignore
+SILKY_MAX_REQUEST_BYTE_SIZE = -1  # Silk takes anything <0 as no limit
+SILKY_MAX_RESPONSE_BYTE_SIZE = 1024  # If response body>1024 bytes, ignore
```
albertyw, line 430: `SILKY_MAX_REQUEST_BODY_SIZE` is being read within `django-silk`: https://github.com/search?q=repo%3Ajazzband%2Fdjango-silk%20SILKY_MAX_RESPONSE_BODY_SIZE&type=code

The variable cannot be changed to `SILKY_MAX_RESPONSE_BYTE_SIZE` in the readme without updating the code within django-silk (which would be a bad idea because it would require users of django-silk to update their code as it would be backwards incompatible)

albertyw (changes requested):

albertyw:
The issue with `BODY`/`BYTES` was fixed in d40bb9dc.  The formatting changes are fine but are separate from what this PR was trying to fix.

PR #700, opened 2024-01-22
Update workflow actions

albertyw:


.github/workflows/test.yml
```diff
@@ -97,7 +97,7 @@ jobs:
     - name: Get pip cache dir
       id: pip-cache
       run: |
-        echo "::set-output name=dir::$(pip cache dir)"
+        echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
```
50-Course, line 100: this change is for better readability?

albertyw: No, just following https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

50-Course: Got it! Thanks for the heads up.

PR #701, opened 2024-01-22
Update test matrix

albertyw:
Test against up-to-date supported versions of Django, Postgres, and MariaDB.


50-Course (approved):
This patch LGTM. Merging it into main trunk

PR #704, opened 2024-02-06
Add support for multipart/form-data requests***

ziyacivan:
**Title:** Add Support for Multipart/Form-Data Requests
**Description**: I've encountered an issue when Silk attempts to handle `multipart/form-data` requests, specifically
when trying to access `request.body` directly. This operation triggers a `RawPostDataException` because Django consumes the stream upon parsing the multipart data, making it inaccessible afterwards.

To address this, I've implemented a check for `multipart/form-data` content type within the request handling logic. When such a content type is detected, the code now properly handles form data and file information using `request.POST` and `request.FILES`, avoiding direct access to `request.body`. This allows Silk to gracefully handle and log multipart requests without raising exceptions.

### DIFF
--- a/silk/model_factory.py
+++ b/silk/model_factory.py
@@ -227,12 +227,12 @@ class RequestModelFactory:
         path = self.request.path
         view_name = self.view_name()
         content_type = self.request.content_type
-        
+
         if content_type.startswith('muştipart/form-data'):
             form_data = self.request.POST.dict()
             files = {f: self.request.FILES[f].name for f in self.request.FILES}
             body = {'form_data': form_data, 'files': files}
-            raw_body = None # Raw body is not available for multipart/form-data
+            raw_body = None  # Raw body is not available for multipart/form-data
         else:
             body, raw_body = self.body()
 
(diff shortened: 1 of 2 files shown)


silk/model_factory.py
```diff
@@ -226,6 +226,16 @@ def construct_request_model(self):
...
+        content_type = self.request.content_type
+
+        if content_type.startswith('multipart/form-data'):
+            body = {
+                'form_data': self.request.POST.dict(),
+                'files': {f: self.request.FILES[f].name for f in self.request.FILES}
+            }
+            raw_body = None  # Raw body is not available for multipart/form-data
+        else:
+            body, raw_body = self.body()
```
50-Course, line 238: @ziyacivan Good catch, and thank you for making this patch! 👏🏼

50-Course (approved):

silk/model_factory.py
ziyacivan: Oh thank you so much! I couldn't continue to the development process (means is tests) because of my social life issues, thank you again 👏🏽

50-Course:
Just found out six different test are failing, I do have to fix them by tomorrow.
By the way, you are welcome to update the test cases if you want to.

Failure is related to implementation, not the test - most likely as a result of writing
implementation before writing the test cases.

PR #707, opened 2024-03-05
Fix test failures on Windows

SpecLad:
`_get_prof_file_name` did not work correctly when the file separator was not `/`.


50-Course (approved):
```python
-        filename = file_path.rsplit('/')[-1]
+        filename = os.path.basename(file_path)
```

LGTM

PR #725, opened 2024-08-12
Correct units in profiling.html and requests.html

### LINKED ISSUE #115: Filters are in milliseconds but indicates seconds
> Request
> Took longer than  **seconds  seconds**, executed more than  n
> queries, and spent longer than **seconds  seconds** executing queries.

ka28kumar:
Fixes #115 

Corrected the units from seconds to milliseconds.

### DIFF
--- a/silk/templates/silk/profiling.html
+++ b/silk/templates/silk/profiling.html
@@ -74,7 +74,7 @@
                    value="{{ filters.overalltime.value }}"/>
             <span style="display:none"></span>
         </div>
-        seconds,
+        milliseconds,
         executed more than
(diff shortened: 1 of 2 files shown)


albertyw (changes requested):
Can you also replace the other references to "seconds" in these files under `placeholder` and for `timespentfilter` with "milliseconds"?

ka28kumar:
Hello @albertyw 
Updated other references. I'm unsure about `filter-seconds-typ`. Can you take a look and let me know if these changes suffice?

albertyw (approved):

PR #768, opened 2025-01-16
Avoid timeout caused by Request with large body

quertenmont:
I am frequently facing timeout because of request with very large body,
These are caused by body processing in request details to generate curl and client commands that are not so essential in most cases (in particular when working with large request)

### DIFF
--- a/silk/views/request_detail.py
+++ b/silk/views/request_detail.py
@@ -26,21 +26,21 @@ class RequestView(View):
             'request': request
         }
 
-        if len(silk_request.raw_body<20000): # Don't do this for large request
+        if len(silk_request.raw_body < 20000):  # Don't do this for large request
             body = silk_request.raw_body
             try:
                 body = json.loads(body)  # Incase encoded as JSON
             except (ValueError, TypeError):
-                pass            
+                pass
             context['curl'] = curl_cmd(url=request.build_absolute_uri(silk_request.path),
-                             method=silk_request.method,
-                             query_params=query_params,
-                             body=body,
-                             content_type=silk_request.content_type),
+                                       method=silk_request.method,
+                                       query_params=query_params,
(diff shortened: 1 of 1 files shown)


albertyw (approved):
Notes:
- Attempting `json.loads` on all requests is going to be computationally heavy (though the code has been doing it for a while)
- `request.html` both have existence checks for `curl` and `client` so they should be fine if this `if` condition is not executed.  It would be nice to have tests for both cases (<20000 and >20000) to prevent regression.
- This is a semi backwards incompatible change in that some users' requests will stop showing curl and client requests.  It would be more elegant for some explanatory text to appear rather than simply the sections in the html output just disappearing.

albertyw (changes requested):
Requesting changes pending passing tests

quertenmont:
Hello @albertyw 
The test which are failing the PR have clearly nothing to do with my changes, so I am not quite sure what I should do to make this PR move on

can you give some suggestions please ?

Loic,

silk/views/request_detail.py
```diff
@@ -19,24 +19,28 @@ def get(self, request, request_id):
...
             'query_params': json.dumps(query_params, sort_keys=True, indent=4) if query_params else None,
-            'client': gen(path=silk_request.path,
-                          method=silk_request.method,
-                          query_params=query_params,
-                          data=body,
-                          content_type=silk_request.content_type),
             'request': request
         }
+
+        if len(silk_request.raw_body < 20000):  # Don't do this for large request
```
albertyw, line 29:
```suggestion
        if len(silk_request.raw_body) < 20000:  # Don't do this for large request
```

quertenmont:
Oups, my bad.
Thanks for spotting this @albertyw

albertyw (approved):

PR #770, opened 2025-01-20
feat: delete all requests using raw_delete to avoid memory issues

moonsikpark:


albertyw:
Is this safe from side effects?  According to [documentation](https://github.com/django/django/blob/main/django/db/models/query.py#L1208), `_raw_delete` doesn't support signals and has no protection for cascades.  Django-silk doesn't utilize signals (users of django-silk might, though unlikely) but there are many objects that do use cascade (`Response`, `SQLQuery`, and `BaseProfile`).

PR #782, opened 2025-04-22
Make autopep8 optional

### LINKED ISSUE #780: Make autopep8 optional
django-silk depends on autopep8 for code formatting.

autopep8 depends on pycodestyle.

Some editors and tooling such as pylsp automatically picks and uses pycodestyle if it is available in the environment.

In our project, we are not using pycodestyle, so we suddenly get a lot of warnings from pycodestyle.

Since the formatting is not a critical feature of django-silk, I propose that to make it optional. If autopep8 is available, it can be used for formatting, otherwise it can avoid the formatting altogether.

This can be achieved with a conditional import like `try: import autopep 8; except ImportError: ...`

A similar issue was brought up before: https://github.com/jazzband/django-silk/issues/660

albertyw:
Fixes #780 

Mark autopep8 as an optional dependency.  If the dependency is available, use it, otherwise return the code as is.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,9 +38,11 @@ setup(
     install_requires=[
         'Django>=4.2',
         'sqlparse',
-        'autopep8',
         'gprof2dot>=2017.09.19',
     ],
+    extras_require={
+        'formatting': ['autopep8'],
+    },
     python_requires='>=3.9',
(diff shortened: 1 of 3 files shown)


aehlke:
please do a release. the latest release breaks on start when autopep8 is missing

PR #788, opened 2025-05-14
Improving SQL details navigation UX

joaopedroalbq:
### Summary
Small "quality of life" improvement for the navigation on the SQL page. This enables the user to open new tabs with their middle mouse click and keeps the previous behavior for the left click.

There is also another minor adjustment to move the logic for the sql detail url into a template tag, cleaning up a small part of the code and addressing a very old TODO comment.

### Context
When debugging a request with multiple queries, it’s frustrating not being able to open each SQL query in its own tab.

Currently, you must open a query, then navigate back each time. This is really inefficient for someone who uses this constantly.

### DIFF
--- a/silk/static/silk/js/pages/sql.js
+++ b/silk/static/silk/js/pages/sql.js
@@ -1,0 +1,17 @@
+$(document).ready(function () {
+  document.querySelectorAll(".data-row").forEach((rowElement) => {
+    let sqlDetailUrl = rowElement.dataset.sqlDetailUrl;
+    rowElement.addEventListener("mousedown", (e) => {
+      switch (e.button) {
+        case 0:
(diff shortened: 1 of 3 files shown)


silk/static/silk/js/pages/sql.js
```diff
@@ -0,0 +1,17 @@
+$(document).ready(function () {
+  document.querySelectorAll(".data-row").forEach((rowElement) => {
+    let sqlDetailUrl = rowElement.dataset.sqlDetailUrl;
+    rowElement.addEventListener("mousedown", (e) => {
```
albertyw, line 4: think this would be better as a "click" event rather than "mousedown", for example in case someone was trying to highlight text.

PR #797, opened 2025-07-17
fix: curl/client values not rendering in request_detail

bcmyguest:
This was a regression from https://github.com/jazzband/django-silk/commit/0a84caaec72ea5ff7739599a64467b1867cd2f2a / https://github.com/jazzband/django-silk/pull/768

This PR fixes a bug where the Curl and Django Test Client code sections were empty. The reason for the bug was that the variables were tuples instead of strings. When the template is rendered, the `curl.strip` method is called which expected a string.

The result of this change is the following:

image

Prior to the change:

image

### DIFF
--- a/silk/views/request_detail.py
+++ b/silk/views/request_detail.py
@@ -36,11 +36,11 @@ class RequestView(View):
                                        method=silk_request.method,
                                        query_params=query_params,
                                        body=body,
-                                       content_type=silk_request.content_type),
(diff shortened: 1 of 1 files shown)


bcmyguest:
Hey @albertyw , Any chance you could take a look at this (or point me to the right maintainer to ask)?

albertyw (approved):

Issue #800, opened 2025-08-06
bug: Usage of transaction.atomic is incorrect when when using non-default database alias


OscarVanL:
# Background

In my application, I use a non-default database alias along with a router.

```
    DATABASES = {
        'default': {
            'ENGINE': '',
        },
        'db1': {
            'ENGINE': 'django.db.backends.mysql',
            'NAME': REDACTED,
            'USER': REDACTED,
            'PASSWORD': REDACTED,
            'HOST': REDACTED,
            'PORT': REDACTED,
        },
        'db2': {
            'ENGINE': 'django.db.backends.mysql',
            'NAME': REDACTED,
            'USER': REDACTED,
            'PASSWORD': REDACTED,
            'HOST': REDACTED,
            'PORT': REDACTED,
        },
    }

    DATABASE_ROUTERS = ['db_router.MyRouter']
```

I have a custom router that decides which database each of my application's `app` models go to, which is based off [this django docs page](https://docs.djangoproject.com/en/4.2/topics/db/multi-db/#automatic-database-routing):
```
class MyRouter:
    APP_DATABASE_MAPPING = {
        "admin": "db1",
        "auditlog": "db1",
        "auth": "db1",
        "contenttypes": "db1",
        "sessions": "db1",
        "silk": "db2",  # django-silk
        "axes": "db2",  # django-axes
    }

    def db_for_read(self, model: models.Model, **hints):
        db = self.APP_DATABASE_MAPPING.get(model._meta.app_label)
        if not db:
            raise ImproperlyConfigured(f"No database configured for {model._meta.app_label}")

        logging.debug(f"Returning DB={db} for {model._meta.object_name} READ")
        return db

    def db_for_write(self, model: models.Model, **hints):
        db = self.APP_DATABASE_MAPPING.get(model._meta.app_label)
        if not db:
            raise ImproperlyConfigured(f"No database configured for {model._meta.app_label}")

        logging.debug(f"Returning DB={db} for {model._meta.object_name} WRITE")
        return db
```

Unfortunately, when I make an API request, this fails in `django-silk` with the following exception:
```
Traceback (most recent call last):
  File "/service/venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/silk/middleware.py", line 96, in __call__
    response = self.process_response(request, response)
[...]

PR #801, opened 2025-08-06
fix: Make transactions target the DB alias selected by the router

OscarVanL:
In https://github.com/jazzband/django-silk/issues/800 I identified a mistake in the transaction handling in the package. This meant that transactions were always started for the `default` DB alias, even if the `DATABASE_ROUTER` used by the project would actually result in silk models being written to a different DB alias.

In my scenario, it caused my application to crash as I do not have a configuration set for the `default` alias.

The fix I have applied is similar to one I found in https://github.com/jazzband/django-axes/pull/1109, where same same bug was once present in `django-axes` (this is another package I use in my project where I do not encounter this bug).

# Assumptions

I have used `router.db_for_write(models.SQLQuery)` to determine the DB alias to use for the transaction. I picked `models.SQLQuery` at random.

This makes an assumption that all of silk's models are routed to the same DB alias. I think this is a fair assumption, but _technically_ someone could make a weird setup where some silk models are routed to one alias, and other models to another alias. This approach does not handle this.

# Testing

I spent far too long trying to write some test cases for this change, but it was just far too difficult. Django didn't seem to permit using `override_settings` to override the `DATABASES` config for a specific test class/case. In the end it proved too challenging, so I have not produced any tests.

I think the change is sufficiently low risk that perhaps we can accept this omission.

### DIFF
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -2,12 +2,13 @@ import logging
 import random
 
 from django.conf import settings
-from django.db import DatabaseError, transaction
+from django.db import DatabaseError, transaction, router
 from django.db.models.sql.compiler import SQLCompiler
 from django.urls import NoReverseMatch, reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
+from silk import models
 from silk.collector import DataCollector
 from silk.config import SilkyConfig
 from silk.errors import SilkNotConfigured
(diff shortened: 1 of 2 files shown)


silk/middleware.py
```diff
@@ -142,28 +143,29 @@ def process_request(self, request):
         request_model = RequestModelFactory(request).construct_request_model()
         DataCollector().configure(request_model, should_profile=should_profile)
 
-    @transaction.atomic()
```
albertyw, line 145: Looking at the `transaction.atomic` implementation in https://github.com/albertyw/django/blob/master/django/db/transaction.py#L316, this looks like it can be modified into `transaction.atomic(using=router.db_for_write(models.SQLQuery))` instead of converting into a `with` block

Other examples: https://sourcegraph.com/search?q=context%3Aglobal+%40transaction.atomic%28using&patternType=keyword&sm=0&df=%5B%22lang%22%2C%22Python%22%2C%22lang%3Apython%22%5D&__cc=1

albertyw (approved):

PR #798, opened 2025-07-18
Fix #665 and #714, add cleanup in middleware

glennmatthews:
Fix #665 and fix #714 by adding try/except logic for the case where the SQL query params contain non-unicode binary data. The observed behavior with this fix is that queries involving such data will not render "explain" content in django-silk, which I think is an acceptable compromise for now versus trying to actually fix the logic here to correctly handle binary data.

I noticed as a side effect of the above bug that (with a long lived server process), django-silk's wrapped `execute_sql` function was being applied to requests that shouldn't be processed by django-silk (i.e. `_should_intercept()` returns False). This is because the wrapper wasn't being removed after processing a request, so all subsequent requests received by the process after any one `_should_intercept() == True` request were still getting the wrapper code. I've fixed this as well by adding appropriate logic in `process_response()` to revert the function wrapper when done with a request.

PR #821, opened 2025-09-07
Skip serialization on explain query

### LINKED ISSUE #819: Getting `django.db.utils.DataError: invalid input syntax for type json` in django-silk versions 5.4.2 and 5.4.1
Getting this error with django-silk 5.4.2 and 5.4.1 (everything was working correctly when I was on 5.4.0):

Example stacktrace:

```
...
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/models/query.py", line 604, in count
    return self.query.get_count(using=self.db)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/models/sql/query.py", line 644, in get_count
    return obj.get_aggregation(using, {"__count": Count("*")})["__count"]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/models/sql/query.py", line 626, in get_aggregation
    result = compiler.execute_sql(SINGLE)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/silk/sql.py", line 104, in execute_sql
    query_dict['analysis'] = _explain_query(self.connection, q, params)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/silk/sql.py", line 66, in _explain_query
    cur.execute(prefixed_query, tuple(force_str_with_fallback(param) for param in params))
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/sentry_sdk/utils.py", line 1816, in runner
    return sentry_patched_function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/sentry_sdk/integrations/django/__init__.py", line 652, in execute
    result = real_execute(self, sql, params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 79, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 92, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 100, in _execute
    with self.db.wrap_database_errors:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/workspace/backend/venv/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
django.db.utils.DataError: invalid input syntax for type json
LINE 1: ...49440f7' AND "hd_git_branches"."data_elements" @> 'Jsonb([{'...
                                                             ^
DETAIL:  Token "Jsonb" is invalid.
CONTEXT:  JSON data, line 1: Jsonb...
```

Happens when I do a `contains` query on a JSONField like this:

```
from django.db.models import Q

Q(my_json_field__contains=[{"foo": "bar"}])
```

In case it helps I was using `psycopg==3.2.9` and `django==5.2.5`

albertyw:
Fixes #819 
Reverts #798 , #807 

@glennmatthews the `_explain_query` switched from using the raw `params` to string params as of #798 .  Since the `params` here are only used for executing the query and are not later saved to a DB like in `execute_sql`, I'm not sure why `params` needed to be `force_str`ed?  As in #819, some database fields are binary and either cannot be decoded (fixed in the try/catch) or should not be decoded even if they can be decoded.


silk/sql.py
```diff
@@ -59,18 +59,7 @@ def _explain_query(connection, q, params):
...
-                Logger.error(
-                    "UnicodeDecodeError while trying to explain query: %s. "
-                    "This could be caused by a non-UTF-8 encoded parameter.",
-                    q
-                )
-                return None
-            cur.execute(prefixed_query, params_str)
+            cur.execute(prefixed_query, params)
             result = _unpack_explanation(cur.fetchall())
             return '\n'.join(result)
```
glennmatthews, line 64: It might be that *these* lines are actually the part that's problematic for non-UTF-8 data? Looking at #798 and specifically the initial commit of 584a6e2fe36e3d8234dc30b45f376f440d4dde38, it was wrapping all three lines here in the try/except at first.

albertyw: Looking at #819 , the error people are getting seem to be an error from the postgresql database itself.  That would only happen in the `cur.execute()` so the exception happens before these two lines are even reached.

PR #847, opened 2026-01-22
Fix mouse event for sql details navigation

albertyw:
As a followup to https://github.com/jazzband/django-silk/pull/788#discussion_r2627321603, fix the event listener to also trigger on middle clicks.

@joaopedroalbq


malefice:
Could anyone please merge this MR? Being able to open multiple detail pages in a quick and efficient way from this list is kind of important.

For everyone else, until this gets merged, you could work around another minor annoyance by not using the back button within the web app (the arrow beside "SQL Query" in my screenshot). You could instead use the browser's back button to just jump back to the SQL list view without triggering a potentially heavy HTTP request. 

image

PR #887, opened 2026-07-23
Fix FiltersManager.get() raising AttributeError without session support

### LINKED ISSUE #361: Empty 'filters' dictonary passed to html renderer
While opening the summary page, i keep getting 'attribute error' as the 'filters' dictionary in the context object is empty. Upon some digging, i see the filters dictionary is set based on 'raw_filters' which in turn is set as below -
raw_filters = request.session.get(self.filters_key, {})
I could see the session object does not contain this information and neither could i see any code that would have set the session object.

agu2347:
Fixes #361.

`get()` falls back to `request.silk_filters` when the request has no `.session` attribute. That attribute is only ever set by `save()` on the same request object, so for a request being processed for the first time (or when session support isn't available at all), it doesn't exist yet, and `get()` raises `AttributeError` instead of returning a default -- unlike the session-based branch just above it, which already safely defaults to `{}` via `request.session.get(self.key, {})`.

This breaks the summary/requests views' filter handling (e.g. context-building code that unconditionally does `raw_filters.pop(...)` and `raw_filters.items()` on the result), matching the reported symptom.

**Fix**: use `getattr(request, 'silk_filters', {})` so `get()` honors the same "always returns a dict, never raises" contract already established by the session branch.

**Testing**: verified directly: constructing a bare object with neither `.session` nor `.silk_filters` and calling `FiltersManager.get()` on it reproduces the exact `AttributeError` with the original code, and returns `{}` correctly with the fix. Confirmed `save()` then `get()` on the same session-less request still round-trips correctly, and that the existing session-based path is completely unaffected.

Added regression tests covering both cases. I confirmed the "no session, no prior save" test fails with the original code and passes with the fix. Ran the full existing `test_filters.py` suite (19 passed: 17 baseline + 2 new) and the summary view tests (2 passed), no regressions.

### DIFF
--- a/project/tests/test_filters.py
+++ b/project/tests/test_filters.py
@@ -232,3 +232,39 @@ class TestProfileFilters(TestCase):
         filtered = query_set.filter(time_taken_filter)
         for f in filtered:
             self.assertGreaterEqual(f.time_taken, c)
+
+
+class TestFiltersManager(TestCase):
+    """
+    Regression tests for
+    https://github.com/jazzband/django-silk/issues/361
+    """
+
+    def test_get_without_session_or_prior_save_returns_empty_dict(self):
+        """
+        get() must never raise, even for a request that has neither a
+        .session attribute nor a previously-set .silk_filters
+        attribute (i.e. save() was never called for this request) --
+        matching the safe default already used by the session-based
+        branch.
+        """
+        from silk.request_filters import FiltersManager
+
+        class BareRequest:
+            pass
+
+        manager = FiltersManager("summary_filters")
+        result = manager.get(BareRequest())
+        self.assertEqual(result, {})
+
+    def test_save_then_get_without_session_round_trips(self):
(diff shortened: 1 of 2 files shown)


project/tests/test_filters.py
```diff
@@ -232,3 +232,38 @@ def test_time_spent_filter(self):
...
+
+    def test_get_without_session_or_prior_save_returns_empty_dict(self):
+        """
+        get() must never raise, even for a request that has neither a
+        .session attribute nor a previously-set .silk_filters
+        attribute (i.e. save() was never called for this request) --
+        matching the safe default already used by the session-based
+        branch.
+        """
+        from silk.request_filters import FiltersManager
```
albertyw, line 251: PEP8: https://peps.python.org/pep-0008/#imports

```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
```

albertyw (approved):

PR #888, opened 2026-07-23
Fix SILKY_IGNORE_PATHS matching when used with a SCRIPT_NAME prefix

### LINKED ISSUE #349: SILKY_IGNORE_PATHS should ignore path without SCRIPT_NAME path prefix
The `settings.SILKY_IGNORE_PATHS` (undocumented) config ignore paths 
https://github.com/jazzband/django-silk/blob/4418b29d23bff2a34edb4310a6472dcf2805eb83/silk/middleware.py#L47

If specified `settings.SILKY_IGNORE_PATHS` (undocumented) matches the `request.path` exactly, it will be excluded from recording. However, since it uses `request.path` as the matching condition, the path will have the path prefix passed from the web front server (apache/nginx). This may or may not match depending on the web front server configuration.

I think django-silk should use `request.path_info` to avoid this problem.
https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.path_info

agu2347:
`SILKY_IGNORE_PATHS` matching now uses `request.path_info` instead of `request.path`. `SILKY_IGNORE_PATHS` is a list of strings written in `settings.py` and not passed through `reverse()`, so they carry no `SCRIPT_NAME` prefix and must be matched against the prefix-stripped `path_info`.

Note: people may have worked around this bug by prefixing their `SILKY_IGNORE_PATHS` with a `SCRIPT_NAME` prefix.  After this is change is released, they will have to remove the prefixes from their `SILKY_IGNORE_PATHS`.

The below text was from the original pull request description:

Fixes #349.

`_should_intercept()` used `request.path` for both: (1) detecting whether a request is for silk's own UI (to avoid silk profiling itself), and (2) checking a configured `SILKY_IGNORE_PATHS` entry. `request.path` includes any SCRIPT_NAME prefix a front-end web server adds ahead of the path Django itself sees, while `get_fpath()` (via `reverse()`) and a user's `SILKY_IGNORE_PATHS` entries are naturally written without that deployment-specific prefix. When silk is deployed behind such a prefix, both checks can therefore fail to match paths they should, causing silk to profile its own UI requests and to ignore the SCRIPT_NAME-prefix mismatch for `SILKY_IGNORE_PATHS` entries -- matching the issue's report (which specifically called out the ignore-paths case, but the same root cause affects the silk-self-detection check too).

`request.path_info` is the SCRIPT_NAME-stripped equivalent -- the same convention Django's own URL resolver uses internally -- and is unaffected by front-end prefix configuration, exactly as the reporter suggested.

**Testing**: verified with a real Django request built via `RequestFactory` with an explicit `SCRIPT_NAME`, against a realistic urlconf with silk mounted under a sub-path (matching typical real-world usage, e.g. `path('silk/', include('silk.urls'))`): confirmed both checks (recognizing silk's own summary page and a configured ignore path) incorrectly failed to match with the original `request.path`-based code, and correctly match with the fix. Also confirmed the common case with no SCRIPT_NAME prefix at all is completely unaffected.

The existing test suite's `TestShouldIntercept` tests construct a bare `silk.models.Request` object (silk's own DB model, not a real `HttpRequest`) and manually set only `.path` on it as a stand-in. Since a real `HttpRequest` always has both `.path` and `.path_info` together, I updated these existing tests to also set `.path_info` (matching what a genuine request would have), and added a new regression test that exercises a `path`/`path_info` mismatch directly, matching the scenario a SCRIPT_NAME prefix produces.

Ran the full existing `test_silky_middleware.py` suite (13 passed: 12 baseline + 1 new). Confirmed the new regression test fails with the original `request.path`-based code and passes with the fix.

### DIFF
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -55,11 +55,20 @@ def _should_intercept(request):
             return False
 
     try:
-        silky = request.path.startswith(get_fpath())
+        # Use path_info (not path) so this comparison is unaffected by
+        # any SCRIPT_NAME prefix a front-end web server may add ahead
+        # of the path Django itself sees -- get_fpath() (via reverse())
+        # returns a path without that prefix, so comparing against
+        # request.path here could incorrectly fail to recognize silk's
+        # own requests when deployed behind such a prefix. See GH #349.
+        silky = request.path_info.startswith(get_fpath())
     except NoReverseMatch:
         silky = False
 
-    ignored = request.path in config.SILKY_IGNORE_PATHS
+    # Likewise, use path_info here so a configured SILKY_IGNORE_PATHS
+    # entry (naturally written without any deployment-specific
+    # SCRIPT_NAME prefix) can still match correctly.
+    ignored = request.path_info in config.SILKY_IGNORE_PATHS
     return not (silky or ignored)
 
 
--- a/silk/middleware.py
+++ b/silk/middleware.py
@@ -55,19 +55,21 @@ def _should_intercept(request):
             return False
 
     try:
-        # Use path_info (not path) so this comparison is unaffected by
-        # any SCRIPT_NAME prefix a front-end web server may add ahead
-        # of the path Django itself sees -- get_fpath() (via reverse())
-        # returns a path without that prefix, so comparing against
-        # request.path here could incorrectly fail to recognize silk's
-        # own requests when deployed behind such a prefix. See GH #349.
-        silky = request.path_info.startswith(get_fpath())
+        # get_fpath() (via reverse()) picks up Django's script prefix
+        # (set from SCRIPT_NAME by the WSGI handler on every real
+        # request), so it already includes any front-end web server
+        # prefix -- exactly like request.path does. Keep comparing
+        # against request.path here; request.path_info has that
+        # prefix stripped and would no longer match.
+        silky = request.path.startswith(get_fpath())
     except NoReverseMatch:
         silky = False
 
-    # Likewise, use path_info here so a configured SILKY_IGNORE_PATHS
-    # entry (naturally written without any deployment-specific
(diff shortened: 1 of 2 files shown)


silk/middleware.py
```diff
@@ -55,11 +55,20 @@ def _should_intercept(request):
...
 
     try:
-        silky = request.path.startswith(get_fpath())
+        # Use path_info (not path) so this comparison is unaffected by
+        # any SCRIPT_NAME prefix a front-end web server may add ahead
+        # of the path Django itself sees -- get_fpath() (via reverse())
+        # returns a path without that prefix, so comparing against
+        # request.path here could incorrectly fail to recognize silk's
+        # own requests when deployed behind such a prefix. See GH #349.
+        silky = request.path_info.startswith(get_fpath())
```
albertyw: I believe `reverse` is already prefixed, so this line should not need to change.

agu2347:
@albertyw You're right, thank you for catching this. I verified it empirically: with `set_script_prefix()`/`SCRIPT_NAME` set the way Django's WSGI handler does on every real request, `reverse()` picks up the script prefix, so `get_fpath()` already includes it -- exactly like `request.path` does. `request.path_info` has it stripped and no longer matches.

I've pushed a correction that reverts this line back to `request.path` and keeps only the `SILKY_IGNORE_PATHS` comparison on `request.path_info` (since those entries are plain strings the user writes in settings.py, never passed through `reverse()`, so they naturally lack the prefix -- that's the actual case #349 reported). I also replaced the old regression test for this line with one that uses `set_script_prefix()` to genuinely simulate a prefixed deployment; confirmed it fails against the old `path_info`-based code and passes with the fix. Full `test_silky_middleware.py` suite: 14 passed.

albertyw (approved):

PR #900, opened 2026-08-13
fix: decompress gzip response bodies in raw view

### LINKED ISSUE #845: @gzip_page breaks raw response view
When a response is provided and the view uses the @gzip_page decorator, even if GZipMiddleware appears *before* silk (as suggested by bug#43), the raw view at `/silk/request/<uuid>/raw/?typ=response&subtyp=raw` will only show the binary gzip data.

Perhaps there should be an option to allow decompressing responses that have `Content-Encoding: gzip` or similar?

Sanjays2402:
Closes #845

Responses compressed by Django's `@gzip_page` decorator were stored with their `Content-Encoding: gzip` header, but Silk's raw response view only base64-decoded the payload and displayed binary gzip data. Decompress gzip-encoded payloads before displaying them while preserving existing behavior for other responses.

Tested with `DB_ENGINE=sqlite3 pytest project/tests/test_models.py`: 68 passed. The regression test fails without the fix and passes with it.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/silk/models.py
+++ b/silk/models.py
@@ -1,4 +1,5 @@
 import base64
+import gzip
 import json
 import random
 import re
@@ -221,7 +222,10 @@ class Response(models.Model):
 
     @property
     def raw_body_decoded(self):
-        return base64.b64decode(self.raw_body)
+        raw_body = base64.b64decode(self.raw_body)
+        if self.headers.get('content-encoding') == 'gzip':
+            return gzip.decompress(raw_body)
+        return raw_body
 
 
 # TODO rewrite docstring
--- a/project/tests/test_models.py
+++ b/project/tests/test_models.py
@@ -1,4 +1,7 @@
+import base64
 import datetime
+import gzip
+import json
 import uuid
 
 from django.core.management import call_command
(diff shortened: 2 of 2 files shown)


silk/models.py
```diff
@@ -221,7 +222,10 @@ def headers(self):
 
     @property
     def raw_body_decoded(self):
-        return base64.b64decode(self.raw_body)
+        raw_body = base64.b64decode(self.raw_body)
+        if self.headers.get('content-encoding') == 'gzip':
+            return gzip.decompress(raw_body)
```
albertyw, line 227: While unlikely, it is possible for `gzip.decompress` to raise an exception for a malformed `raw_body`.  This should probably be wrapped in a try/except to prevent it from raising an uncaught exception.

albertyw (approved):

Sanjays2402:
Thanks @albertyw — good catch. Wrapped the `gzip.decompress` call in `raw_body_decoded` in a try/except (`OSError`, `EOFError`), so a malformed `raw_body` falls back to the still-compressed body instead of 500ing the raw view. Added a test covering the malformed case alongside the existing happy-path one.

albertyw (approved):

PR #905, opened 2026-08-21
Preserve in-flight requests when clearing silk logs

### LINKED ISSUE #824: IntegrityError when clearing request log with in-flight request
I occasionally see `FOREIGN KEY constraint failed` errors coming from silk; my hypothesis is that it's caused by the following race condition:

1. a request comes into django
2. silk creates a Request record
3. `silk_clear_request_log` runs, deleting all Request records
4. django is done with the request
5. silk tries to insert a Response record referencing the Request record from step 2
6. IntegrityError is raised as step 3 deleted the Request that the Response is referencing

For now, I'm working around the issue with my own cleanup command that skips deleting any in-flight requests, but perhaps it's something that could be handled directly by `silk_clear_request_log`?

richardmilles:
Preserve in-flight requests (no `end_time`) when clearing or garbage-collecting silk logs so response finalization cannot hit a foreign-key IntegrityError.
Fixes #824

### DIFF
--- a/silk/utils/data_deletion.py
+++ b/silk/utils/data_deletion.py
@@ -26,3 +26,29 @@ def delete_model(model):
         if not items_to_delete:
             break
         model.objects.filter(pk__in=items_to_delete).delete()
+
+
+def clear_silk_data():
+    """Delete silk request logs, preserving in-flight requests.
+
+    In-flight requests have ``end_time`` unset. Clearing them races with
+    response finalization and can raise IntegrityError when silk inserts a
+    Response for a Request that was just deleted.
+    """
+    # Imported lazily to avoid circular imports at module load.
+    from silk.models import Profile, Request, Response, SQLQuery
+
+    in_flight_ids = list(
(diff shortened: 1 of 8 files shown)


silk/utils/data_deletion.py
```diff
@@ -26,3 +26,29 @@ def delete_model(model):
...
+    response finalization and can raise IntegrityError when silk inserts a
+    Response for a Request that was just deleted.
+    """
+    # Imported lazily to avoid circular imports at module load.
+    from silk.models import Profile, Request, Response, SQLQuery
+
+    in_flight_ids = list(
+        Request.objects.filter(end_time__isnull=True).values_list('pk', flat=True)
+    )
+    if not in_flight_ids:
```
albertyw, line 44: If this command were run on a live database (i.e. which would likely be a database with in-flight requests), wouldn't there be a race condition where 1) `in_flight_ids` is empty, 2) a request gets started, 3) `delete_model` gets called?  Should there be additional exception handling?

